PR #34, opened 2021-04-23
Multi-Layer-Component

KaroFr:
- Removed initialAmount so the two inputs remain once you submit.
- Removed layerSubscription and added updateLayersForSelection() so the selected layers remain the same once you submit.
- Removed updateLayer because it caused an unnecessary connection between the layers selected. Changing the first selected layer sometimes caused a change in the second selected layer too.


KaroFr:
[Expression hat nur noch 1x Input, nachdem man sie abgeschickt hat](https://trello.com/c/ohYjikHb/25-expression-hat-nur-noch-1x-input-nachdem-man-sie-abgeschickt-hat)

PR #42, opened 2021-04-29
update time also on zero

ChristianBeilschmidt:


ChristianBeilschmidt:
[Time-Selection: Reset button not enabled when changing seconds from 0 to 1](https://trello.com/c/UVgTNZ1R/144-time-selection-reset-button-not-enabled-when-changing-seconds-from-0-to-1)

PR #45, opened 2021-05-03
allow nan as no data and check value for validity

michaelmattig:


michaelmattig:
[Expression: unable to set NaN as nodata for float](https://trello.com/c/E6kx4TpN/35-expression-unable-to-set-nan-as-nodata-for-float)

projects/wave-core-new/src/lib/operators/dialogs/expression-operator/expression-operator.component.html
```diff
@@ -41,7 +41,7 @@
 
         <mat-form-field>
             <mat-label>No-Data Value</mat-label>
-            <input matInput type="number" formControlName="noDataValue" />
+            <input matInput type="text" formControlName="noDataValue" />
```
ChristianBeilschmidt, line 44: Why text?

projects/wave-core-new/src/lib/operators/dialogs/expression-operator/expression-operator.component.ts
```diff
@@ -152,6 +153,43 @@ export class ExpressionOperatorComponent implements AfterViewInit, OnDestroy {
         // this.unitSubscription.unsubscribe();
     }
 
+    validateNoData(control: AbstractControl): ValidationErrors | null {
+        if (!isNaN(parseFloat(control.value)) || control.value.toLowerCase() === 'nan') {
```
ChristianBeilschmidt, line 157: You can express no data as a number, but just not with JSON.
IMHO you should work with number here and as potential string for submitting it to our server.

ChristianBeilschmidt (changes requested):

projects/wave-core-new/src/lib/operators/dialogs/expression-operator/expression-operator.component.html
michaelmattig: s.t. users can set NAN as the no-data value. That was the whole point of the task... Now they can either write a number or "nan" in order to create the layer

projects/wave-core-new/src/lib/operators/dialogs/expression-operator/expression-operator.component.ts
michaelmattig: users have to be able to set NAN as the selected no data value

projects/wave-core-new/src/lib/operators/dialogs/expression-operator/expression-operator.component.html
ChristianBeilschmidt: Hm, I see. We could allow NAN only for floats and we should then present a checkbox for NAN.
This box could disable the number field then.
What do you think?

michaelmattig: that would also be possible. however the way I did it is also fine imho.

PR #67, opened 2021-06-17
tag enums with type

michaelmattig:


michaelmattig:
[API enums anders taggen](https://trello.com/c/fdqe0BLV/106-api-enums-anders-taggen)

projects/wave-core-new/src/lib/backend/backend.model.ts
```diff
@@ -354,11 +366,10 @@ export interface MetaDataSuggestionDict {
     metaData: MetaDataDefinitionDict;
 }
 
-export interface MetaDataDefinitionDict {
-    OgrMetaData?: OgrMetaDataDict;
-}
+export type MetaDataDefinitionDict = OgrMetaDataDict;
 
 export interface OgrMetaDataDict {
+    type: 'ogrMetaData';
```
ChristianBeilschmidt, line 372: Maybe we should allow a capital beginning for Operators and Types…

projects/wave-core-new/src/lib/datasets/dataset.model.ts
```diff
@@ -75,18 +77,19 @@ export class InternalDatasetId {
...
+            type: 'internal',
+            dataset: this.id,
         };
     }
 }
 
 export class ExternalDatasetId {
     provider: UUID;
-    id: string;
+    dataset: string;
```
ChristianBeilschmidt, line 88:
```suggestion
    id: string;
```

projects/wave-core-new/src/lib/layers/symbology/symbology.model.ts
```diff
@@ -90,12 +88,12 @@ export abstract class Symbology {
 
 export abstract class VectorSymbology extends Symbology {
     static fromVectorSymbologyDict(dict: VectorSymbologyDict): VectorSymbology {
```
ChristianBeilschmidt, line 90: I guess there is not such thing as a vector symbology dict anymore? cf. other PR

projects/wave-core-new/src/lib/operators/dialogs/point-in-polygon-filter/point-in-polygon-filter.component.ts
```diff
@@ -65,9 +65,25 @@ export class PointInPolygonFilterOperatorComponent {
...
+                                            },
+                                            stroke: {
+                                                width: {
+                                                    type: 'static',
+                                                    value: 1,
+                                                },
+                                                color: {
+                                                    type: 'static',
+                                                    color: [0, 0, 0, 255],
+                                                },
```
ChristianBeilschmidt, line 81: We should really think about having the type for static stuff always here. I guess it's good for programming logic but looks very weird.

ChristianBeilschmidt (changes requested):

projects/wave-core-new/src/lib/backend/backend.model.ts
michaelmattig: everything I changed is a type. So you want `"type": "LinearGradient"` and `"type": "Internal"` capitalized?

projects/wave-core-new/src/lib/datasets/dataset.model.ts
michaelmattig: I think `dataset` makes more sense here, because `povider` is also actually an id. Plus, the id is stored in an `id` attribute so it would be like `id.id`

projects/wave-core-new/src/lib/layers/symbology/symbology.model.ts
michaelmattig: yeah there is no such dict but there is still a VectorSymbology class that encapsulates common behavior which is good

projects/wave-core-new/src/lib/operators/dialogs/point-in-polygon-filter/point-in-polygon-filter.component.ts
michaelmattig: it's not pretty but its constistent to always have the type

projects/wave-core-new/src/lib/datasets/dataset.model.ts
ChristianBeilschmidt: now you have 
```{
  "dataset": {
    "dataset": …
  }
}
```

ChristianBeilschmidt (approved):

PR #73, opened 2021-06-29
fix type tags

jdroenner:


jdroenner:
[fix geoengine-ui main type tags](https://trello.com/c/vD1W9sn5/242-fix-geoengine-ui-main-type-tags)

PR #78, opened 2021-07-11
Tests for project service

KaroFr:


KaroFr:
[Methoden des ProjectService testen: *project](https://trello.com/c/51OGCznh/38-methoden-des-projectservice-testen-project)

PR #79, opened 2021-07-11
update  node in ci

jdroenner:


jdroenner:
[update node in geoengine-ui ci](https://trello.com/c/3CwgxMOM/261-update-node-in-geoengine-ui-ci)

PR #81, opened 2021-07-24
polygons as join input

jdroenner:


jdroenner:
[polygons as join input](https://trello.com/c/OCWWyCH9/275-polygons-as-join-input)

PR #87, opened 2021-08-04
angular-v12

ChristianBeilschmidt:


angular.json
```diff
@@ -94,7 +95,13 @@
                             "node_modules/codemirror/lib/codemirror.css",
                             "node_modules/ol/ol.css"
                         ],
-                        "scripts": []
+                        "scripts": [],
+                        "vendorChunk": true,
+                        "extractLicenses": false,
+                        "buildOptimizer": false,
+                        "sourceMap": true,
+                        "optimization": false,
```
michaelmattig, line 103: why no optimization?

ChristianBeilschmidt: As far as I see this is false for the default context (dev) but set to true in the **production** case.

michaelmattig (approved):

PR #127, opened 2021-10-19
handle disabled anonymous sessions

michaelmattig:


projects/wave-core-new/src/lib/users/modal-login/modal-login.component.html
```diff
@@ -0,0 +1,11 @@
+<form [formGroup]="loginForm" (ngSubmit)="login()">
+    <p>Anonymous usage is disabled. Please log in.</p>
```
jdroenner, line 2: remove "Anonymous usage is disabled."

jdroenner (changes requested):

projects/wave-core-new/src/lib/users/modal-login/modal-login.component.html
```diff
@@ -0,0 +1,11 @@
+<form [formGroup]="loginForm" (ngSubmit)="login()">
```
ChristianBeilschmidt, line 1: Shouldn't we add a header here?

For other things like the lineage popup (https://github.com/geo-engine/geoengine-ui/blob/main/projects/wave-core-new/src/lib/provenance/lineage-graph/lineage-graph.component.html) and the rename popup (https://github.com/geo-engine/geoengine-ui/blob/main/projects/wave-core-new/src/lib/layers/rename-layer/rename-layer.component.html), there is always one.

ChristianBeilschmidt (changes requested):
Johannes said that the `undefined` on top in the time navigation looks like a bug / an error in our application. Maybe we can just hide it if it is undefined? Or show a blank space or a `'-'` :thinking:.

ChristianBeilschmidt (approved):

jdroenner (approved):

PR #159, opened 2021-12-01
only call workflow methds for current session and once

ChristianBeilschmidt:
The problem was that within the get-*-workflow methods was a call to the session stream. But the stream would be updated and so a new request would be made. This would fail, if the new session cannot access the data of the previous user.

To fix this, I used the `sessionTokenForRequest` method that only emits ONCE.


ChristianBeilschmidt:
[Falsches Project wird upgedatet bei User-Wechsel](https://trello.com/c/wcvQNBOF/473-falsches-project-wird-upgedatet-bei-user-wechsel)

michaelmattig (approved):

michaelmattig:
bors merge

PR #231, opened 2022-03-03
add plot to ebv analyzer

michaelmattig:


michaelmattig:
[Zeit-Plot wieder einbinden](https://trello.com/c/sPo1Czgx/653-zeit-plot-wieder-einbinden)

projects/ebv-analyzer-app/src/app/ebv-selector/ebv-selector.component.html
```diff
@@ -122,23 +122,18 @@
...
-                                 [xLimits]="plotSettings.xLimits"
-                                 [yLabel]="plotSettings.yLabel"
-                                 [yLimits]="plotSettings.yLimits"></wave-ebv-indicator-plot> -->
+    <div fxLayout="row" fxLayoutAlign="space-between start">
+        <wave-ebv-country-selector fxFlex=""></wave-ebv-country-selector>
+        <span fxFlex="1rem"></span>
+        <button mat-raised-button color="primary" [disabled]="isPlotButtonDisabled$ | async" (click)="plot()">Calculate</button>
     </div>
 
+    <wave-vega-viewer *ngIf="(plotData | async) !== undefined" [chartData]="plotData | async"></wave-vega-viewer>
```
ChristianBeilschmidt, line 133: Didn't you have your trick of binding to the output of ngIf and using that in the other field?

projects/ebv-analyzer-app/src/app/ebv-selector/ebv-selector.component.ts
```diff
@@ -209,6 +213,92 @@ export class EbvSelectorComponent implements OnInit, OnDestroy {
         return element.name.toLowerCase().includes(filterString);
     }
 
+    plot(): void {
+        combineLatest([
+            this.dataSelectionService.rasterLayer.pipe(
+                mergeMap<RasterLayer | undefined, Observable<RasterLayer>>((layer) => (layer ? of(layer) : of())),
```
ChristianBeilschmidt, line 219: `of()` is the same as `EMPTY`, this never yields. Is that the intended purpose?

projects/ebv-analyzer-app/src/app/ebv-selector/ebv-selector.component.ts
```diff
@@ -612,3 +566,15 @@ function guessDataRange(colorizer: Colorizer): {min: number; max: number} {
 
     return {min, max};
 }
+
+// TODO: use method from core
```
ChristianBeilschmidt, line 570: why don't you use it? I guess it is already there

projects/ebv-analyzer-app/src/app/country-provider.service.ts
```diff
@@ -36,93 +35,53 @@ export class CountryProviderService {
...
+                                    },
+                                },
+                                fillColor: {
+                                    type: 'static',
+                                    color: [0, 0, 0, 0],
+                                },
+                            }),
+                            isLegendVisible: false,
+                            isVisible: true,
+                        }),
```
ChristianBeilschmidt, line 78: Two things:
* I guess we could use GEO BON green and maybe some very transparent filling?
* Maybe we can define it independently from this function

projects/ebv-analyzer-app/src/app/ebv-selector/ebv-selector.component.html
```diff
@@ -122,23 +122,18 @@
...
-        <!-- TODO: implement -->
-        <!-- <wave-ebv-indicator-plot *ngIf="plotSettings"
-                                 [data]="plotSettings.data$"
-                                 [xLimits]="plotSettings.xLimits"
-                                 [yLabel]="plotSettings.yLabel"
-                                 [yLimits]="plotSettings.yLimits"></wave-ebv-indicator-plot> -->
+    <div fxLayout="row" fxLayoutAlign="space-between start">
+        <wave-ebv-country-selector fxFlex=""></wave-ebv-country-selector>
+        <span fxFlex="1rem"></span>
+        <button mat-raised-button color="primary" [disabled]="isPlotButtonDisabled$ | async" (click)="plot()">Calculate</button>
```
ChristianBeilschmidt, line 130: In the current EBV Analyzer, we only show this after we selected an entity. I guess it makes sense, since we only then allow some computation.

ChristianBeilschmidt: You can also already click calculate and nothing happens.

projects/ebv-analyzer-app/src/app/ebv-selector/ebv-selector.component.html
```diff
@@ -122,23 +122,18 @@
...
+    <mat-divider></mat-divider>
 
-        <!-- TODO: implement -->
-        <!-- <wave-ebv-indicator-plot *ngIf="plotSettings"
-                                 [data]="plotSettings.data$"
-                                 [xLimits]="plotSettings.xLimits"
-                                 [yLabel]="plotSettings.yLabel"
-                                 [yLimits]="plotSettings.yLimits"></wave-ebv-indicator-plot> -->
+    <div fxLayout="row" fxLayoutAlign="space-between start">
+        <wave-ebv-country-selector fxFlex=""></wave-ebv-country-selector>
```
ChristianBeilschmidt, line 128: when I select first Germany and then Czechia, Germany is still on the map

projects/ebv-analyzer-app/src/app/ebv-selector/ebv-selector.component.ts
```diff
@@ -209,6 +213,92 @@ export class EbvSelectorComponent implements OnInit, OnDestroy {
...
+                        workflowId,
+                        {
+                            time: {
+                                start: (this.ebvTree as EbvHierarchy).tree.time.start,
+                                end: (this.ebvTree as EbvHierarchy).tree.time.end,
+                            },
+                            bbox: extentToBboxDict(viewport.extent),
+                            crs: crs.srsString,
+                            // TODO: set reasonable size
+                            spatialResolution: [0.1, 0.1],
```
ChristianBeilschmidt, line 283: Do we need to set this based on our dataset? Or on our bounds?

ChristianBeilschmidt (changes requested):

projects/ebv-analyzer-app/src/app/ebv-selector/ebv-selector.component.html
michaelmattig: What trick? I copied this snipped from the plots in the others portals

projects/ebv-analyzer-app/src/app/ebv-selector/ebv-selector.component.ts
michaelmattig: the method is not exported and I don't know how to do it.

projects/ebv-analyzer-app/src/app/country-provider.service.ts
michaelmattig: in the current portal the outline is also red.

projects/ebv-analyzer-app/src/app/ebv-selector/ebv-selector.component.ts
michaelmattig: I am not sure. The plot isn't responsive to the map so we should just use some reasonable numbers here

projects/ebv-analyzer-app/src/app/ebv-selector/ebv-selector.component.html
michaelmattig: 268112d718dfd89b7fa17a05dbfe6748221c3b48

projects/ebv-analyzer-app/src/app/ebv-selector/ebv-selector.component.html
michaelmattig: I can't reproduce that, they switch fine for me. Do you get any error?

projects/ebv-analyzer-app/src/app/ebv-selector/ebv-selector.component.ts
ChristianBeilschmidt: You have to add the file where it is in into the `public-api.ts`.
I guess it is just a model, so you can sort it in there.
Please keep the sorting for a better overview.

projects/ebv-analyzer-app/src/app/ebv-selector/ebv-selector.component.html
ChristianBeilschmidt: You bounded the async call to a variable and could use the variable then without having to subscribe twice. You introduced this concept…

projects/ebv-analyzer-app/src/app/ebv-selector/ebv-selector.component.ts
ChristianBeilschmidt: Could you use a filter then instead? If not, we can keep it as is.

projects/ebv-analyzer-app/src/app/ebv-selector/ebv-selector.component.ts
ChristianBeilschmidt: I guess we could do it in the original resolution of the dataset. I'm not sure if it is slow then.

projects/ebv-analyzer-app/src/app/ebv-selector/ebv-selector.component.html
ChristianBeilschmidt: Now it works. Didn't work before.

ChristianBeilschmidt:
Please merge the current master

projects/ebv-analyzer-app/src/app/ebv-selector/ebv-selector.component.ts
michaelmattig: I copied this code and I think it uses a map instead of a filter because it changes the type from `RasterLayer | undefined` to `RasterLayer` which I guess a filter does not.

projects/ebv-analyzer-app/src/app/ebv-selector/ebv-selector.component.ts
michaelmattig: the public-api.ts only exports Services, Components, Directives and Models but not plain functions. That's why I don't know how to use those in another project.

projects/ebv-analyzer-app/src/app/ebv-selector/ebv-selector.component.html
michaelmattig: f79b05467731ed2d842d28424e885a8f6cd4dd58

projects/ebv-analyzer-app/src/app/ebv-selector/ebv-selector.component.ts
ChristianBeilschmidt: You can either put it under models or misc.
For instance, `form.validators.ts` has `isValidUuid` which you can import since it is `export`ed.
Or `rgbaColorFromDict` from `color.ts`.

michaelmattig: how do I import it then from a different project? both functions you mention are only used in the core

ChristianBeilschmidt: You import them the same way you import components, directives and so on.

`import {rgbaColorFromDict} from 'wave-core';`

michaelmattig: when I first tried that it said that it doesn't know that component but it works
7697d8327a582eb6fec3147830fff733f91de6ab

ChristianBeilschmidt (approved):

ChristianBeilschmidt:
bors merge

ChristianBeilschmidt:
bors merge

PR #287, opened 2022-04-09
Column Range Filter

KaroFr:


KaroFr:
[UI für ColumnRangeFilter](https://trello.com/c/xCkEbhn1/679-ui-f%C3%BCr-columnrangefilter)

PR #361, opened 2022-06-23
EBV selector can now work with regular and irregular time steps

ChristianBeilschmidt:
I needed to model the enum for regular time series or irregular time series as list.


jdroenner:
bors merge

PR #397, opened 2022-08-08
Add Open ID Connect login component

glombiewski:
Open ID Connect login component. No unit tests yet.


glombiewski:
[[FAIR-DS / RESPECT 6.4] Authentication and Authorization Infrastructure (AAI)](https://trello.com/c/vPV8PSeO/769-fair-ds-respect-64-authentication-and-authorization-infrastructure-aai)

projects/gfbio-app/src/app/oidc/oidc.component.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    ngOnInit(): void {
+        this.userSubscription = this.userService
+            .getSessionStream()
+            .pipe(first())
+            .subscribe((session) => {
+                if (!session.user || session.user.isGuest) {
+                } else {
+                    this.user = session.user;
+                }
```
ChristianBeilschmidt, line 31: It is either missing something, or you mean:

```suggestion
                if (!session.user || session.user.isGuest) {
                    return;
                }
                
                this.user = session.user;
```

projects/gfbio-app/src/app/oidc/oidc.component.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    constructor(
+        @Inject(Config) private readonly config: AppConfig,
+        private readonly userService: UserService,) {
+    }
+
+    ngOnInit(): void {
+        this.userSubscription = this.userService
+            .getSessionStream()
+            .pipe(first())
```
ChristianBeilschmidt, line 26: If you use `first`, then this subscription terminates after the first value.
This means you don't need to save it and destroy it at a later point.

projects/gfbio-app/src/app/oidc/oidc.component.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    login() : void {
+        this.loginSubscription = this.userService.oidcInit().subscribe((idr) => {
+                window.location.href = idr.url;
+            }
+        );
+    }
+
+    logout() : void {
+        this.logoutSubscription = this.userService.guestLogin().pipe(first()).subscribe({
```
ChristianBeilschmidt, line 43: Same here

projects/gfbio-app/src/app/oidc/oidc.component.ts
```diff
@@ -0,0 +1,68 @@
...
+            .subscribe((session) => {
+                if (!session.user || session.user.isGuest) {
+                } else {
+                    this.user = session.user;
+                }
+            });
+    }
+
+    login() : void {
+        this.loginSubscription = this.userService.oidcInit().subscribe((idr) => {
```
ChristianBeilschmidt, line 36: and maybe here

ChristianBeilschmidt: Especially if you allow calling this function and thus creating it several times, you would either need to destroy it at the header of the function… or you don't need to unsubscribe after all by using something like `first`, etc.

projects/wave-core-new/src/lib/users/user.service.ts
```diff
@@ -207,6 +207,26 @@ export class UserService {
 
         return of(session);
     }
+
+    oidcInit() : Observable<AuthCodeRequestURL> {
+        return this.backend.oidcInit();
+    }
```
ChristianBeilschmidt, line 213: This call is what you want to avoid, right? You had some TODO in the backend PR.

projects/wave-core-new/src/lib/users/user.service.ts
```diff
@@ -207,6 +207,26 @@ export class UserService {
...
+            .subscribe(
+                (session) => {
+                    this.session$.next(session);
+                    result.next(session);
+                },
+                (error) => result.error(error),
+                () => result.complete(),
+            );
+        return result.asObservable();
+    }
```
ChristianBeilschmidt, line 229: I guess you need a replaySubject in order to prevent race-conditions.
I've created a helper function here (https://github.com/geo-engine/geoengine-ui/blob/944a795aeb32cc96da765ea35c7bd78e966608f5/projects/wave-core-new/src/lib/project/project.service.ts#L847-L857) because we ran into problems.

But it could be that this is only the case if this emits multiple values :thinking: .

Maybe we can refactor this into a util function and use it also here.

ChristianBeilschmidt (changes requested):

projects/gfbio-app/src/app/oidc/oidc.component.ts
glombiewski: Reasoning:
If a user opens the login component and clicks the login button, the backend (for whatever reason) might take a moment to respond. If the user closes the login component, and starts doing something else, whenever the backend returns the call, the user will still be redirected to the OIDC provider.

I would expect that to be weird behavior, because I closed the login component. There are some discussions about it online, but I defer to your preference.

projects/gfbio-app/src/app/app.component.ts
```diff
@@ -233,6 +239,7 @@ export class AppComponent implements OnInit, AfterViewInit {
      * @return true, if the splash dialog should be skipped, false otherwise
      */
     private handleQueryParameters(): void {
+        this.tryLogin();
```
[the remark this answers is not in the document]
glombiewski: I need an explanation what the baskets even are to think about this properly.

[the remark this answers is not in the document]
ChristianBeilschmidt: Baskets are GFBio items that are submitted to Geo Engine UI via a URL parameter.
The app will load these items and append them as layers.
If the layer list is not empty, it will show a popup that shows several options.

projects/gfbio-app/src/app/oidc/oidc.component.ts
ChristianBeilschmidt: Okay, I see.

projects/gfbio-app/src/app/app.component.ts
```diff
@@ -189,6 +190,11 @@ export class AppComponent implements OnInit, AfterViewInit {
                 icon: 'help',
                 tooltip: 'Help',
             },
+            {
+                sidenavConfig: {component: OidcComponent},
+                icon: 'contacts',
+                tooltip: 'Login',
+            },
```
ChristianBeilschmidt, line 197: We should not put this below help. Maybe this can be on top.

projects/gfbio-app/src/app/oidc/oidc.component.html
```diff
@@ -0,0 +1,11 @@
...
+<ng-container *waveIfLoggedIn>
+    <div *ngIf="user">Welcome {{ user.realName }}!</div>
+    <button mat-raised-button color="primary" (click)="logout()">Logout</button>
+</ng-container>
+<ng-container *waveIfGuest>
+    <div>Please login</div>
+    <div *ngIf="user === undefined">Undefined</div>
+    <div *ngIf="user !== undefined && user.isGuest">Guest User</div>
+    <button mat-raised-button color="primary" (click)="login()">Login with GFBio</button>
+</ng-container>
```
ChristianBeilschmidt, line 11: Maybe, when you add some margin between "Welcome…" and the button, this component looks okay (80:20 wise).

It should not say
```
Please login
Undefined
```

Maybe add a sentence that you will be redirected to a SSO site; and hide the undefined variable.

projects/gfbio-app/src/app/oidc/oidc.component.scss
```diff
@@ -0,0 +1,9 @@
+:host {
+    display: block;
+    padding: 1rem;
+}
+button {
+    width: 20%;
+    margin-left: 40%;
+    margin-right: 40%;
```
ChristianBeilschmidt, line 8: I guess you need a min-width in `rem` of the button, so that "Logout" fits in there.
Moreover, with `marign: 0 auto;` you can center stuff.

ChristianBeilschmidt (changes requested):

projects/gfbio-app/src/app/app.component.ts
```diff
@@ -244,4 +251,28 @@ export class AppComponent implements OnInit, AfterViewInit {
...
+                            console.log(session);
+                        },
+                        (e) => {
+                            console.log(e); //TODO: Remove for final version
+                        },
+                    );
+            }
+        } else {
+            //TODO: Probably no else necessary.
+        }
```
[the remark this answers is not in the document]
glombiewski: 015443e2f1feb5b36f066f65977c31238e677cba

projects/gfbio-app/src/app/app.module.ts
```diff
@@ -26,9 +26,10 @@ import {HelpComponent} from './help/help.component';
 import {SplashDialogComponent} from './splash-dialog/splash-dialog.component';
 import {BasketDialogComponent} from './basket/basket-dialog/basket-dialog.component';
 import {BasketService} from './basket/basket.service';
+import {OidcComponent} from "./oidc/oidc.component";
 
 @NgModule({
-    declarations: [AppComponent, HelpComponent, SplashDialogComponent, BasketDialogComponent],
+    declarations: [AppComponent, HelpComponent, SplashDialogComponent, BasketDialogComponent, OidcComponent],
```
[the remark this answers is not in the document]
glombiewski: dfd1dac5eab19917b548030c8211d8c565a5cc96

projects/gfbio-app/src/app/app.component.ts
glombiewski: dfd1dac5eab19917b548030c8211d8c565a5cc96

projects/gfbio-app/src/app/oidc/oidc.component.ts
glombiewski: I disabled the button while a request is pending. A BehaviorSubject seemed like the best way, but if there is a better one, let me know. c136f37f0518115ce6a71ddb29722e6b0cb70703 28a3bec946cc25c3de364287db08fa2edd3646c5

projects/wave-core-new/src/lib/users/user.service.ts
glombiewski: I don't think this call can be avoided. It would be nice to get rid of the tryLogin call in the AppComponent by sending an answer directly to the backend, but I don't know how the backend would get the frontend back.

projects/gfbio-app/src/app/oidc/oidc.component.html
glombiewski: d0e929eb0b9cc0d3e987c9dd16658a4479d3f7c8

projects/gfbio-app/src/app/oidc/oidc.component.scss
glombiewski: I looked over other button usages. Sometimes they are aligned right (e.g. Boxplot), sometimes aligned left (Basic Statistics) and sometimes they are width 100% (Save As in Workspace Settings). 

I don't know if there is a reasoning behind the button placements and widths, but the last one seemed the most similar use case to this, so I styled this in the same manner. d0e929eb0b9cc0d3e987c9dd16658a4479d3f7c8

projects/wave-core-new/src/lib/users/user.service.ts
glombiewski: I honestly did not think about this, since it is the same for the normal login. FWIW, I don't see the race condition.

projects/gfbio-app/src/app/oidc/oidc.component.ts
```diff
@@ -0,0 +1,68 @@
...
+import {AppConfig} from "../app-config.service";
+import {first, SubscriptionLike} from "rxjs";
+
+@Component({
+    selector: 'wave-gfbio-login',
+    templateUrl: 'oidc.component.html',
+    styleUrls: ['./oidc.component.scss'],
+})
+export class OidcComponent implements OnInit, OnDestroy {
+    private userSubscription : SubscriptionLike | undefined;
```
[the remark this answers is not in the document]
glombiewski: 989c122b2d2d8284a699dd4d04cd5e277c912dcb

projects/gfbio-app/src/app/app.component.ts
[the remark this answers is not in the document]
glombiewski: I guess this is correct: 989c122b2d2d8284a699dd4d04cd5e277c912dcb ?

projects/wave-core-new/src/lib/users/user.service.ts
ChristianBeilschmidt: I guess then it is also wrong for the normal login.

The race is:

1. create the subject
2. set next
3. set completed
4. subscribe to the  subject
5. get completed

…and I guess you would expect to get the session in `next`.
It happens rarely but caused a bug in the project service some months ago.

With the replay subject, the `next` value would be replayed. But you can correct me if I am wrong here.

glombiewski: Ok. I guess the issue with the refactoring is session$.next, which is not part of the subscribeAndProvide function.

Does the order of this.session$.next(session) and result.next(session) matter? E.g., the IfLoggedInDirective triggers based on the session$, so I guess there might also be an issue if it is updated and result isn't (or vice versa). In practice, I would expect this to resolve itself - but clearly, I don't yet fully understand when those rxjs events trigger.

glombiewski: Added a ReplaySubject for now. 3f6c5b249eddc2a067ffa5c5ffd717cf3b8f0bcc

ChristianBeilschmidt (approved):

ChristianBeilschmidt:
bors merge

PR #407, opened 2022-08-12
Column range fix

ClockworkOnion:
Added a check so lexicography min < max check is only performed on text columns


michaelmattig (approved):

michaelmattig:
bors merge

PR #158, opened 2021-11-25
float values for raster legend

KaroFr:

PR #461, opened 2022-09-21
Legend display improvements

ClockworkOnion:
- Measurement displayed above data ("NDVI" in screenshot below)
- Decimals rounded to 3 places after the last  continuing zero, eg 34.0012345 becomes 34.00123

image


ClockworkOnion:
[UI-Ästhetik für TimeSlider und Legende](https://trello.com/c/ImESjfsC/982-ui-%C3%A4sthetik-f%C3%BCr-timeslider-und-legende)

projects/core/src/lib/layers/legend/legend-raster/raster-legend.component.ts
```diff
@@ -87,4 +87,17 @@ export class RasterLegendComponent implements OnInit, OnChanges {
...
+        const integerPart: number = Math.floor(toShorten);
+        const decimals: number = toShorten - integerPart;
+        const decimalsString = decimals.toString();
+        let zeroes = 0;
+        while (decimalsString.charAt(zeroes + 2) === '0') {
+            zeroes++;
+        }
+        const roundAt: number = Math.pow(10, zeroes + DECIMAL_PLACES);
+        return integerPart + Math.round(decimals * roundAt) / roundAt;
+    }
```
ChristianBeilschmidt, line 102: I don't think this is a good heuristic for the use case.
You have to consider all values that you want to display.
Then, when you want to distinguish them, you need to add decimal places to do so.

Also, you want to be kind of consistent with all the numbers to once specify one decimal place and other times six.

I was wondering now if we did not tackle this problem before and did some research.
Please take a look at https://github.com/geo-engine/geoengine-ui/pull/158/files.
I'm not quite sure where this code went.

Then, like Karo did back then, please add a test for different arrays of breakpoints to see that they are all calculated properly.

ChristianBeilschmidt (changes requested):

projects/core/src/lib/layers/legend/legend-raster/raster-legend.component.ts
ClockworkOnion: Here is what I came up with:
There might be a case when values differ only in their last few digits, but are otherwise digits other than zero, for example 3.1111112345 and 3.1111119876. This probably happens very rarely, but in theory it might happen. So right now I'm first comparing all the values to see if there is any overlap. If so, rounding must take place 3 decimal places behind the last index where all values overlap. Otherwise I'm checking for the greatest amount of zeroes, which then becomes the minimum amount of places for all numbers, before any rounding takes place.
The larger of these two values takes precedence.

A worst case scenario would be when two values are like the above examples, but the others are all entirely different values (4, 5 etc), so those two would be flattened to 3.111... but I guess that can't happen since breakpoints are evenly distributed?

The code still needs some cleanup and refactoring, and I still need to add the tests (tested my methods in a separate environments for simplicity) but perhaps you could give me some feedback on whether I'm on the right track.

ChristianBeilschmidt: In general, you can cut off after the last digit which helps you to distinguish between the values.

If you have significant gaps, for instance, 1,000,000.23345435 to 5,000,000.4365464, you can, in principle, cut all decimal places. On the other hand, if you have values distributed between 3.1111112345 and 3.1111119876, you must not make them indistinguishable by outputting [3, 3].

As stated above, please write some tests for different cases so that you can see the effects.

projects/core/src/lib/layers/legend/legend-raster/raster-legend.component.html
```diff
@@ -1,21 +1,22 @@
 <tbody class="legend">
     <ng-template [ngIf]="layer.symbology.colorizer.isContinuous()">
-        <tr *ngFor="let breakpoint of layer.symbology.colorizer.getBreakpoints(); let isFirst = first">
+        <!-- measurement -->
+        <ng-container *ngIf="measurement$ | async | continuousMeasurement as measurement">
+            ( {{ measurement.measurement }} <span *ngIf="measurement.unit">in {{ measurement.unit }}</span> )
+        </ng-container>
+        <ng-container *ngIf="measurement$ | async | classificationMeasurement as measurement">
+            ( {{ measurement.measurement }} )
```
ChristianBeilschmidt, line 8: 1. Now you've put everything in braces which looks bad. Please only put the unit part in braces.

grafik

2. There should not be whitespace before and after the brace.

3. Moreover, please add a small margin to the bottom so that it looks nice.

projects/core/src/lib/layers/legend/legend-raster/raster-legend.component.spec.ts
```diff
@@ -46,4 +46,48 @@ describe('RasterLegend', () => {
         ];
         expect(calculateNumberPipeParameters(breakpoints6)).toBe('1.0-0');
     });
+
+    it('unifyDecimals', () => {
```
ChristianBeilschmidt, line 50: There is an error running this test in the CI.
Please first let the CI run through before assigning the PR for review.

projects/core/src/lib/layers/legend/legend-raster/raster-legend.component.ts
```diff
@@ -56,6 +56,52 @@ export class CastMeasurementToContinuousPipe implements PipeTransform {
...
+    let maxOverlap = 0;
+    values.forEach((x) => {
+        values.forEach((y) => {
+            if (x === y) {
+                return;
+            }
+            const overlap: number = overlappingDigits(x, y);
+            maxOverlap = overlap > maxOverlap ? overlap : maxOverlap;
+        });
+    });
```
ChristianBeilschmidt, line 82: This seems to be highly inefficient. Please just check the adjacent values in a single loop.

projects/core/src/lib/layers/legend/legend-raster/raster-legend.component.ts
```diff
@@ -56,6 +56,52 @@ export class CastMeasurementToContinuousPipe implements PipeTransform {
...
+    const str1 = val1.toString();
+    const str2 = val2.toString();
+    const maxLength = Math.min(str1.length, str2.length);
+    let passedDecimal = false;
+    while (overlap < maxLength && str1.charAt(overlap) === str2.charAt(overlap)) {
+        if (str1.charAt(overlap) === '.' && str2.charAt(overlap) === '.') {
+            passedDecimal = true;
+        }
+        overlap++;
+    }
```
ChristianBeilschmidt, line 101: Can't you make a for loop from 0 to maxLength and break early?

projects/core/src/lib/layers/legend/legend-raster/raster-legend.component.ts
```diff
@@ -56,6 +56,52 @@ export class CastMeasurementToContinuousPipe implements PipeTransform {
...
+                return;
+            }
+            const overlap: number = overlappingDigits(x, y);
+            maxOverlap = overlap > maxOverlap ? overlap : maxOverlap;
+        });
+    });
+    return values.map((x) => {
+        const preDecimals = Math.floor(x).toString().length;
+        const roundAt = Math.pow(10, Math.max(0, maxOverlap + 2 - preDecimals));
+        return Math.round(x * roundAt) / roundAt;
```
ChristianBeilschmidt, line 86: What is the rationale for rounding instead of just cutting off?

projects/core/src/lib/layers/legend/legend-raster/raster-legend.component.ts
```diff
@@ -56,6 +56,52 @@ export class CastMeasurementToContinuousPipe implements PipeTransform {
     }
 }
 
+export function unifyDecimals(values: number[]): number[] {
+    // Early return if all values differ by more than 1
+    if (values.every((x) => x >= 1)) {
```
ChristianBeilschmidt, line 61: I guess instead of first checking that everything is >= 1, just do the one loop and check that the values are apart.

projects/core/src/lib/layers/legend/legend-raster/raster-legend.component.ts
```diff
@@ -56,6 +56,52 @@ export class CastMeasurementToContinuousPipe implements PipeTransform {
...
+        const toCompare: number[] = values.map((x) => Math.floor(x)).sort((a, b) => a - b);
+        let oneApart = true;
+        for (let i = 0; i < toCompare.length - 1; i++) {
+            if (toCompare[i + 1] - toCompare[i] < 1) {
+                oneApart = false;
```
ChristianBeilschmidt, line 66: if it is false, you can `break`

ChristianBeilschmidt: why don't you make a helper function that returns a boolean and in that case returns early?

ChristianBeilschmidt (changes requested):

projects/core/src/lib/layers/legend/legend-raster/raster-legend.component.html
ClockworkOnion: I moved the braces to the unit, deleted whitespace and put 0.5rem as a bottom margin. 330864ceb13c28994b236530b8cf6744a8428b7f

projects/core/src/lib/layers/legend/legend-raster/raster-legend.component.ts
ClockworkOnion: Right, in hindsight, what I did there is pretty embarrassing. Not sure why I thought I had to compare every value with every other one. I condensed it to a single loop that checks the adjacent values.

projects/core/src/lib/layers/legend/legend-raster/raster-legend.component.ts
ClockworkOnion: Changed the while-loop to a for-loop with an early break if any difference is found. 330864ceb13c28994b236530b8cf6744a8428b7f

projects/core/src/lib/layers/legend/legend-raster/raster-legend.component.ts
ClockworkOnion: I thought it would represent the actual number more accurately. But since the goal is primarily to have distinct values, I changed it to Math.floor here (and adjusted the tests accordingly) 330864ceb13c28994b236530b8cf6744a8428b7f

projects/core/src/lib/layers/legend/legend-raster/raster-legend.component.ts
ClockworkOnion: Right, that check was completely unnecessary and a remnant from when I used to only cut off all the decimals when the integer part was at least three digits. I removed the if-check.

projects/core/src/lib/layers/legend/legend-raster/raster-legend.component.ts
ClockworkOnion: Moved the code to a helper function and added an early return 330864ceb13c28994b236530b8cf6744a8428b7f

ChristianBeilschmidt (approved):

ChristianBeilschmidt:
bors merge

PR #465, opened 2022-09-23
Time slider hidden when there is no time to select

ClockworkOnion:
The time slider is hidden when only one timestep exists, i.e. there is nothing to select.

### DIFF
--- a/projects/core/src/lib/time/time-step-selector/time-step-selector.component.html
+++ b/projects/core/src/lib/time/time-step-selector/time-step-selector.component.html
@@ -12,6 +12,6 @@
     <span>{{ currentTimeFormatted | async }}</span>
 </div>
 
-<div *ngIf="!(displayTimeLine())" class="padded container mat-typography" fxLayout="row" fxLayoutAlign="space-between center">
+<div *ngIf="!displayTimeLine()" class="padded container mat-typography" fxLayout="row" fxLayoutAlign="space-between center">
     <span>Point in Time: {{ currentTimeFormatted | async }}</span>
 </div>
--- a/projects/core/src/lib/time/time-step-selector/time-step-selector.component.html
+++ b/projects/core/src/lib/time/time-step-selector/time-step-selector.component.html
@@ -1,4 +1,4 @@
-<div *ngIf="displayTimeLine()" class="container mat-typography" fxLayout="row" fxLayoutAlign="space-between center">
+<div *ngIf="displayTimeSlider()" class="container mat-typography" fxLayout="row" fxLayoutAlign="space-between center">
     <mat-slider
         thumbLabel="true"
         [min]="min"
@@ -12,6 +12,6 @@
     <span>{{ currentTimeFormatted | async }}</span>
 </div>
 
-<div *ngIf="!displayTimeLine()" class="padded mat-typography">
+<div *ngIf="!displayTimeSlider()" class="padded mat-typography">
     <span>{{ currentTimeFormatted | async }}</span>
 </div>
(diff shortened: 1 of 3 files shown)


ClockworkOnion:
[UI-Ästhetik für TimeSlider und Legende](https://trello.com/c/ImESjfsC/982-ui-%C3%A4sthetik-f%C3%BCr-timeslider-und-legende)

projects/core/src/lib/time/time-step-selector/time-step-selector.component.html
```diff
@@ -1,4 +1,4 @@
-<div class="container mat-typography" fxLayout="row" fxLayoutAlign="space-between center">
+<div *ngIf="displayTimeLine()" class="container mat-typography" fxLayout="row" fxLayoutAlign="space-between center">
```
ChristianBeilschmidt, line 1: If you do it like this, the whole time information is gone.

I guess it makes more sense to only hide the slider component but not the time information.

ChristianBeilschmidt (changes requested):

projects/core/src/lib/time/time-step-selector/time-step-selector.component.html
ChristianBeilschmidt: I would just show the year, not "Point in Time" before that. It is the same info as with the slider.

projects/core/src/lib/time/time-step-selector/time-step-selector.component.ts
```diff
@@ -101,4 +101,8 @@ export class TimeStepSelectorComponent implements OnInit, OnChanges, OnDestroy {
             return timeStep.start.format(this.timeFormat);
         };
     }
+
+    displayTimeLine(): boolean {
```
ChristianBeilschmidt, line 105:
```suggestion
    displayTimeSlider(): boolean {
```

projects/core/src/lib/time/time-step-selector/time-step-selector.component.scss
```diff
@@ -6,6 +6,11 @@
     }
 }
 
+.padded {
+    padding: 1rem;
+    padding-left: 2rem;
```
ChristianBeilschmidt, line 11:
```suggestion
    padding: 1rem 1rem 1rem 2rem;
```

Do you know why you have to adjust it more to the left?

projects/core/src/lib/time/time-step-selector/time-step-selector.component.html
```diff
@@ -11,3 +11,7 @@
     ></mat-slider>
     <span>{{ currentTimeFormatted | async }}</span>
 </div>
+
+<div *ngIf="!displayTimeLine()" class="padded container mat-typography" fxLayout="row" fxLayoutAlign="space-between center">
```
ChristianBeilschmidt, line 15: Please try to make it as high as with the slider.

ChristianBeilschmidt (changes requested):

projects/core/src/lib/time/time-step-selector/time-step-selector.component.scss
ClockworkOnion: That's because the `.container` span property has a right margin of 1rem, without which the box would end directly after the timestamp. Is it okay this way, or should I try to reorganize the values so that this right/left offset is unified?

projects/core/src/lib/time/time-step-selector/time-step-selector.component.html
ClockworkOnion: I found that with a padding of `0.85rem` and margin of `0 0.2rem` it's practically identical to how it looks with the slider. I guess these are oddly specific values, but I can't figure out a nicer way to emulate the padding/margin that happens due to the slider, when it is there. d64c6de9ffb5fdb85f1f9cb85c7612b7e4b260dd

projects/core/src/lib/time/time-step-selector/time-step-selector.component.ts
ClockworkOnion: Renamed the method

ChristianBeilschmidt (approved):

ChristianBeilschmidt:
bors merge

PR #492, opened 2022-10-25
cancel wms queries when the results are no longer needed

michaelmattig:


projects/core/src/lib/project/project.service.ts
```diff
@@ -850,6 +850,20 @@ export class ProjectService {
         return this.selectedFeature$.value;
     }
 
+    /**
+     * Get a stream that signals whether a running query should be aborted because the results are no longer needed
+     */
+    getQueryAbortStream(): Observable<any> {
```
ChristianBeilschmidt, line 856: "create" since you can only use it once?

ChristianBeilschmidt, line 856: please return something like `void`, but not `any`

projects/core/src/lib/map/map-layer.component.ts
```diff
@@ -311,10 +311,18 @@ export class OlRasterLayerComponent
 
         this.source.setTileLoadFunction((tile, src) => {
             const client = new XMLHttpRequest();
+
+            const sub = this.projectService.getQueryAbortStream().subscribe((_) => {
+                client.abort();
+                sub.unsubscribe();
+                this.source.dispatchEvent('tileloaderror');
+            });
+
```
ChristianBeilschmidt, line 320: Could there be a race condition in doing this?

ChristianBeilschmidt (changes requested):

projects/core/src/lib/map/map-layer.component.ts
michaelmattig: I think it could happen that the request completes exactly at the same time as we try to abort it. But `abort` does nothing if the request already finished. It could happen that the `tileloadend` and `tileloaderror` both get dispatched. But as the condition for changing the rasterLayerStatus is `<=` we should be fine also. It could happen that we try to unsubscribe the subscription twice, I don't know what happens then, do you?

projects/core/src/lib/project/project.service.ts
michaelmattig: cdcf2d8422c2a68248a0fd29e85970fad57ad68d

michaelmattig: cdcf2d8422c2a68248a0fd29e85970fad57ad68d

ChristianBeilschmidt:
[Peek 2022-10-28 16-56.webm](https://user-images.githubusercontent.com/4947061/198662252-531a13ef-5193-441a-b277-1f13e088d3d3.webm)

I've used NDVI from EPSG:3857. Then I created an expression that uses this 2x, adds both and divides it by two. Then… use it like in the video. Moreover, I don't see cancel logs anymore.

ChristianBeilschmidt (approved):

ChristianBeilschmidt:
bors merge

PR #505, opened 2022-11-04
fix query abort during zoom

michaelmattig:


michaelmattig:
[Raster Tiles fehlen nach Zoomen](https://trello.com/c/xepRLJGH/1078-raster-tiles-fehlen-nach-zoomen)

projects/core/src/lib/map/map-layer.component.ts
```diff
@@ -311,12 +320,16 @@ export class OlRasterLayerComponent
...
-            const sub = this.projectService.createQueryAbortStream().subscribe(() => {
-                (tile as any).getImage().src = 'assets/images/empty_tile.png';
-                client.abort();
-                sub.unsubscribe();
-                this.source.dispatchEvent('tileloaderror');
+            const tileZoomLevel: number = (tile as any).getTileCoord()[0];
+            const tileResolution = this.mapService.getView().getResolutionForZoom(tileZoomLevel);
+
+            const sub = this.projectService.createQueryAbortStream().subscribe((newResolution) => {
+                if (newResolution !== tileResolution) {
```
ChristianBeilschmidt, line 327: Does this mean we do not abort now when the user pans?

We should abort there, too.

projects/core/src/lib/project/project.service.ts
```diff
@@ -863,8 +865,7 @@ export class ProjectService {
 
         return combineLatest(observables).pipe(
             skip(1),
-            take(1),
-            map(() => {}),
+            map((a) => a[1].resolution),
```
ChristianBeilschmidt, line 868: What is **a** and why do we take the resolution of the "second a"?

projects/core/src/lib/map/map-layer.component.ts
```diff
@@ -311,12 +320,16 @@ export class OlRasterLayerComponent
...
+                    (tile as any).setState(TileState.EMPTY);
+                    client.abort();
+                    sub.unsubscribe();
```
ChristianBeilschmidt, line 330: Why do you unsubscribe from yourself?

1. You'll get blind from this.
2. Just make sure that this stream finishes after the first `onNext`. You can achieve this for instance with `first()`.

projects/core/src/lib/map/map-layer.component.ts
```diff
@@ -311,12 +320,16 @@ export class OlRasterLayerComponent
...
+                    this.source.dispatchEvent('tileloadend');
+                }
```
ChristianBeilschmidt, line 332: In the documentation of OpenLayers, it states some slightly different behavior of custom loading, where it sets the error state at error and ok at… ok.

```js
import TileState from 'ol/TileState';

source.setTileLoadFunction(function(tile, src) {
  const xhr = new XMLHttpRequest();
  xhr.responseType = 'blob';
  xhr.addEventListener('loadend', function (evt) {
    const data = this.response;
    if (data !== undefined) {
      tile.getImage().src = URL.createObjectURL(data);
    } else {
      tile.setState(TileState.ERROR);
    }
  });
  xhr.addEventListener('error', function () {
    tile.setState(TileState.ERROR);
  });
  xhr.open('GET', src);
  xhr.send();
});
```

Maybe adapting this helps.

There is also the event `'abort'`. Maybe you can call abort manually if needed but react on all kind of abortions by using the default handler.

https://openlayers.org/en/latest/apidoc/module-ol_Tile.html#~LoadFunction

projects/core/src/lib/map/map-layer.component.ts
```diff
@@ -311,12 +320,16 @@ export class OlRasterLayerComponent
...
         this.source.setTileLoadFunction((tile, src) => {
             const client = new XMLHttpRequest();
-
-            const sub = this.projectService.createQueryAbortStream().subscribe(() => {
-                (tile as any).getImage().src = 'assets/images/empty_tile.png';
-                client.abort();
-                sub.unsubscribe();
-                this.source.dispatchEvent('tileloaderror');
+            const tileZoomLevel: number = (tile as any).getTileCoord()[0];
+            const tileResolution = this.mapService.getView().getResolutionForZoom(tileZoomLevel);
```
ChristianBeilschmidt, line 324: I guess you could use `getTileCoord` instead: https://openlayers.org/en/latest/apidoc/module-ol_Tile-Tile.html

It gives you x, y, and zoom level z.

ChristianBeilschmidt (changes requested):

projects/core/src/lib/project/project.service.ts
michaelmattig: 857ab2e12b8cf1b629860ce691677f01650c8d14 better? please make a suggestion instead of a rhetorical question then.

projects/core/src/lib/map/map-layer.component.ts
michaelmattig: d3de124e84e0164decaf638cfe0aa245d8068a25

projects/core/src/lib/map/map-layer.component.ts
michaelmattig: it's no longer the case that we only need the first event. It may be that the first event is just the delayed zoom event. We need to keep the subscription until we either decide to abort, or the request finishes. That's why I have a unsubscribe here and one in the onload function

projects/core/src/lib/map/map-layer.component.ts
michaelmattig: We use the `tileloadend` event for the loading indicator.

projects/core/src/lib/map/map-layer.component.ts
michaelmattig: I already use it

projects/core/src/lib/map/map-layer.component.ts
ChristianBeilschmidt: As discussed, you can refactor the `createQueryAbortStream` method to incorporate the zoom level as parameter.
Then, you only need to wait for it to push the one and only abort event.

projects/core/src/lib/map/map-layer.component.ts
ChristianBeilschmidt: How does the loading indicator get this xhr object?

Moreover, we could just implement loaded, error and abort and handle the tile stuff there.

In the subscription, we just need to call abort.

projects/core/src/lib/map/map-layer.component.ts
michaelmattig: you mean the resolution and not the zoom level, right? Shall I also include the map extent then? Otherwise we still have to listen for multiple events.

projects/core/src/lib/map/map-layer.component.ts
michaelmattig: I meant that we count the number of `tileloadevent`s etc. to determine the status of the loading indicator. How will we determine that if we no longer dispatch the event?

projects/core/src/lib/map/map-layer.component.ts
ChristianBeilschmidt: Both, you should make a queryAbortStream that actually fires when the query should be aborted.

projects/core/src/lib/map/map-layer.component.ts
ChristianBeilschmidt: I did not say don't dispatch an event if this is necessary.
Just refactor the handling according to the docs and add `abort` if possible.

projects/core/src/lib/map/map-layer.component.ts
michaelmattig: 4f52ff455e47336a8b782b01bf7c826bcf07c9a4

projects/core/src/lib/map/map-layer.component.ts
michaelmattig: I refactored it. I tried setting the TileState to EMPTY in a `addEventListener('abort')` method, but I got a tile status sequence violation error. As we do not set the state anywhere else, I assume this interferes with the internal state update. So I left it out. It works fine for me like this.

ChristianBeilschmidt: I tried around, the thing with the sequence violation error is quite easy to grasp because each state is a number and you must not set a lower state.

I tried this:

```typescript
    // TileState.IDLE = 0
            // TileState.LOADING = 1
            // TileState.LOADED = 2
            // TileState.ERROR = 3
            // TileState.EMPTY = 4

            // tile.setState(TileState.LOADING);

            const debug_id = Math.random();

            client.addEventListener('loadend', (_event) => {
                console.log('loadend', debug_id);
                console.log('[1] tile.getState()', tile.getState());

                cancelSub.unsubscribe();
                const data = client.response;

                if (tile.getState() === TileState.EMPTY) {
                    console.log(data);
                }

                if (tile.getState() === TileState.EMPTY) {
                    return; // tile was aborted
                }

                if (!data) {
                    tile.setState(TileState.ERROR);
                } else {
                    (tile as any).getImage().src = URL.createObjectURL(data);
                    tile.setState(TileState.LOADED);
                }

                console.log('[2] tile.getState()', tile.getState());
            });
            client.addEventListener('error', () => {
                console.log('error', debug_id);
                console.log('tile.getState()', tile.getState());

                tile.setState(TileState.ERROR);
            });
            client.addEventListener('abort', () => {
                console.log('abort', debug_id);
                console.log('tile.getState()', tile.getState());

                tile.setState(TileState.EMPTY);
            });
            client.send();
        });
```

My insights:
1. After abort, there will no matter what be loadend called. Thus, you have to check whether to set ERROR then, because 3 < 4.
2. When you set LOADING at the beginning, our loading indicator will never stop (did not investigate this further)
3. When you set the tile to EMPTY, the loading indicator also won't stop loading (maybe please also check this)
4. For whatever reasons it seems to be fine to set nothing if everything goes fine :shrug: .
5. I guess you have to set TileState.ERROR on ERROR or ABORT. This is what you currently do, without knowing I guess, since the expectation was that we only set ERROR on error.

projects/core/src/lib/map/map-layer.component.ts
```diff
@@ -309,25 +321,37 @@ export class OlRasterLayerComponent
...
+        const proj = olGetProj(this.spatialReference.srsString) as Projection;
+        const tileGrid = this.source.getTileGridForProjection(proj);
+
+        this.source.setTileLoadFunction((olTile, src) => {
+            const tile = olTile as OlImageTile;
+            const tileCoord = tile.getTileCoord();
+            const tileZoomLevel = tileCoord[0];
+            const tileExtent = tileGrid.getTileCoordExtent(tileCoord) as Extent;
+
+            const tileResolution = this.mapService.getView().getResolutionForZoom(tileZoomLevel);
```
ChristianBeilschmidt, line 333: You don't need this here, since you only use this as an input to `createQueryAbortStream`.

But the `ProjectService` already has this as a dependency, so you can take this from there.

This also means you can revert adding the map service to the layer.

ChristianBeilschmidt (changes requested):

projects/core/src/lib/map/map-layer.component.ts
michaelmattig: makes sense. So I tested the loading indicator and it works when tiles are loaded normally, aborted, and errors are thrown. So it seems to me we can resolve this.

projects/core/src/lib/map/map-layer.component.ts
michaelmattig: 4c39c8f4c743670e5368119dd47061314c52246d

projects/core/src/lib/map/map-layer.component.ts
```diff
@@ -309,25 +314,35 @@ export class OlRasterLayerComponent
...
-                (tile as any).getImage().src = data;
-            };
+            client.addEventListener('loadend', (_event) => {
+                cancelSub.unsubscribe();
+                const data = client.response;
+
+                if (!data) {
+                    tile.setState(TileState.ERROR);
+                } else {
+                    (tile as any).getImage().src = URL.createObjectURL(data);
```
ChristianBeilschmidt, line 340:
```suggestion
                    (tile.getImage() as HTMLImageElement).src = URL.createObjectURL(data);
```

ChristianBeilschmidt (approved):

projects/core/src/lib/map/map-layer.component.ts
ChristianBeilschmidt: last thing

ChristianBeilschmidt:
bors d+

ChristianBeilschmidt (approved):

ChristianBeilschmidt:
bors merge

PR #511, opened 2022-11-10
Call node with option to prefer IPv4 DNS results in local development

dbrandenstein:


ChristianBeilschmidt:
1. This unfortunately does not work for me:

```bash
 *  Executing task: npm run serve:gis:local 


> geoengine@0.7.0 serve:gis:local
> node --dns-result-order ipv4first ./node_modules/@angular/cli/bin/ng serve --configuration development gis --proxy-config proxy-local.conf.json

node: bad option: --dns-result-order

 *  The terminal process "/usr/bin/bash '-c', 'npm run serve:gis:local'" failed to launch (exit code: 9). 
 *  Terminal will be reused by tasks, press any key to close it. 
```

This is the version from my Ubuntu 22 LTS:

```bash
$ node --version
v14.17.3
```

2. The calling of node with the path to ng is kind of ugly in one line. Could you add an alias for that? Like a command that wraps it and you call the command?

dbrandenstein:
> This unfortunately does not work for me

Sorry about that. I checked the CLI docs for the Node 14 LTS and it listed the option, but I missed the detail that it was only added in v14.18.0. Since in the readme we only require >= 14.x.x this does not seem to be a good solution then for now, unless we change the requirement. Ubuntu 22.04 LTS actually includes an even lower version v12.22.9, but that should be fine because the readme says to use Node 14+.
Alternatively we could just add a sentence to the readme mentioning the option to set the environment variable if someone is using Node >= 17 and having problems.

ChristianBeilschmidt:
1. Please change the readme so that it says to use Node 16 with its version being >= 16.18.X
2. Please remove 14 from the CI and maybe add a TODO to include 18 there once we upgrade to Angular 15.
3. Please find out if we can have like one script in the package.json that does this parametrization stuff and the other scripts just call this.
    Alternatively, maybe try *pre*-scripts, like `"preserve:gis:local": "NODE_OPTIONS=\"--dns-result-order ipv6first\"",`.

package.json
```diff
@@ -4,20 +4,21 @@
...
         "serve:dashboards/ebv-analyzer:nightly": "ng serve --configuration development ebv-analyzer --proxy-config proxy-ebv-analyzer.conf.json",
-        "serve:dashboards/gfbio:local": "ng serve --configuration development gfbio --proxy-config proxy-local.conf.json",
+        "serve:dashboards/gfbio:local": "ng-prefer-ipv4 serve --configuration development gfbio --proxy-config proxy-local.conf.json",
         "serve:dashboards/gfbio:nightly": "ng serve --configuration development gfbio --proxy-config proxy-gfbio.conf.json",
-        "serve:dashboards/nfdi-gdo:local": "ng serve --configuration development nfdi-gdo --proxy-config proxy-local.conf.json",
+        "serve:dashboards/nfdi-gdo:local": "ng-prefer-ipv4 serve --configuration development nfdi-gdo --proxy-config proxy-local.conf.json",
         "serve:dashboards/nfdi-gdo:nightly": "ng serve --configuration development nfdi-gdo --proxy-config proxy-nfdi-gdo.conf.json",
-        "serve:gis:local": "ng serve --configuration development gis --proxy-config proxy-local.conf.json",
+        "serve:gis:local": "ng-prefer-ipv4 serve --configuration development gis --proxy-config proxy-local.conf.json",
```
ChristianBeilschmidt, line 21: I thought this should work, but it somehow does not:

```
 *  Task wird ausgeführt: npm run serve:gis:local 


> geoengine@0.7.0 serve:gis:local
> ng-prefer-ipv4 serve --configuration development gis --proxy-config proxy-local.conf.json

sh: 1: ng-prefer-ipv4: not found

 *  Der Terminalprozess "/usr/bin/bash '-c', 'npm run serve:gis:local'" konnte nicht gestartet werden (Exitcode: 127). 
 *  Das Terminal wird von Aufgaben wiederverwendet, drücken Sie zum Schließen eine beliebige Taste. 
```

```
node --version
v16.18.1
``

**Review by ChristianBeilschmidt — CHANGES_REQUESTED** [review]

File: package.json (hunk 1 above)
Comment (dbrandenstein) [reply package.json]: Fixed in https://github.com/geo-engine/geoengine-ui/pull/511/commits/7e93b23de08d584ee0e76749bad2acacc823ed8e.
Not sure though whether it's still worth it for the only slightly shorter line :D
I could additionally pull the "serve --proxy-config proxy-local.conf.json" part into the "ng-prefer-ipv4" script, though it becomes kind of hard to comprehend which options are set where then. At least now it's seaparated between node options and angular options.

File: package.json [hunk 2]
```diff
@@ -4,20 +4,21 @@
...
     "scripts": {
         "ng": "ng",
+        "ng-prefer-ipv4": "node --dns-result-order ipv4first ./node_modules/@angular/cli/bin/ng",
         "build": "npm run build-prod:core && npm run build-apps",
         "build:core": "ng build --configuration development core",
         "build-watch:core": "ng build --configuration development core --watch",
         "build-prod:core": "ng build --configuration production core",
         "build-apps": "npm run build-prod:gis && npm run build-prod:dashboards/data-atlas && npm run build-prod:dashboards/ebv-analyzer && npm run build-prod:dashboards/gfbio && npm run build-prod:dashboards/nfdi-gdo",
-        "serve:dashboards/data-atlas:local": "ng serve --configuration development data-atlas --proxy-config proxy-local.conf.json",
+        "serve:dashboards/data-atlas:local": "npm run ng-prefer-ipv4 -- serve --configuration development data-atlas --proxy-config proxy-local.conf.json",
```
ChristianBeilschmidt, line 13: That works :+1: .

I guess we can make this even better:

```
"ng-serve-local": "node --dns-result-order ipv4first ./node_modules/@angular/cli/bin/ng serve --configuration development $1 --proxy-config proxy-local.conf.json",
"serve:gis:local": "npm run ng-serve-local gis",
```

ChristianBeilschmidt (changes requested):

package.json
dbrandenstein: Done in https://github.com/geo-engine/geoengine-ui/pull/511/commits/75ca64dac66cb755d486da9707d0c4d9ac50cf89.
The "$1" does not do anything however, because arguments are always simply appended. So I left it out and changed the order of the arguments in the helper script.
(There exists a workaround which allows for positional arguments but it relies on shell, so leads to issues on Windows cmd apparently.)

ChristianBeilschmidt (approved):

ChristianBeilschmidt:
You have to remove the 14.X stuff from https://github.com/geo-engine/geoengine-ui/blob/main/bors.toml in order for me to merge it.

ChristianBeilschmidt:
bors merge

ChristianBeilschmidt (approved):

ChristianBeilschmidt:
bors merge

PR #513, opened 2022-11-14
Symbology Editor - Global Layer Properties: Apply color changes manually

ClockworkOnion:
- After making changes to the global layer properties inside the symbology editor, the changes are not immediately applied to the map view. Instead, the user gets notified of unapplied changes by the background color changing, the heading changing and the "Apply changes" button becoming active, as seen in attached screenshot. Pressing the button will apply changes to the active layer, return the button to inactive etc.
- "Colorizer Type" buttons now wrap, also seen in screenshot, to prevent some buttons being completely hidden on smaller screens
- Fixed a small typo in "Add Data" -> "Upload" ("Upload data from you local computer" -> "you" to "your")

image

### DIFF
--- a/projects/core/src/lib/colors/color-map-selector/color-map-selector.component.html
+++ b/projects/core/src/lib/colors/color-map-selector/color-map-selector.component.html
@@ -72,8 +72,6 @@
             </table>
         </div>
 
-        <!-- <mat-action-row>
-            <button type="submit" mat-raised-button color="primary" [disabled]="form.invalid">Create color table</button>
-        </mat-action-row> -->
+        <!-- TODO: will be instantiated to populate the color table -->
     </form>
 </div>
--- a/projects/core/src/lib/colors/color-map-selector/color-map-selector.component.html
+++ b/projects/core/src/lib/colors/color-map-selector/color-map-selector.component.html
@@ -72,6 +72,9 @@
             </table>
         </div>
 
-        <!-- TODO: will be instantiated to populate the color table -->
+        <!-- TODO: This will be instantiated to populate the color table -->
+        <!-- <mat-action-row>
+            <button type="submit" mat-raised-button color="primary" [disabled]="form.invalid">Create color table</button>
+        </mat-action-row> -->
     </form>
 </div>
--- a/projects/core/src/lib/colors/color-map-selector/color-map-selector.component.html
+++ b/projects/core/src/lib/colors/color-map-selector/color-map-selector.component.html
@@ -1,5 +1,5 @@
 <div class="component_container">
-    <form [formGroup]="form" (ngSubmit)="applyNewColorTable($event)" fxLayout="column">
+    <form [formGroup]="form" (ngSubmit)="applyChanges()" fxLayout="column">
         <div
             formGroupName="bounds"
             class="flex-container"
@@ -72,8 +72,8 @@
             </table>
         </div>
 
-        <mat-action-row>
+        <!-- <mat-action-row>
             <button type="submit" mat-raised-button color="primary" [disabled]="form.invalid">Create color table</button>
-        </mat-action-row>
+        </mat-action-row> -->
     </form>
 </div>
--- a/projects/core/src/lib/layers/symbology/raster-symbology-editor/raster-symbology-editor.component.html
+++ b/projects/core/src/lib/layers/symbology/raster-symbology-editor/raster-symbology-editor.component.html
@@ -93,7 +93,7 @@
                 </ng-template>
             </mat-expansion-panel>
 
-            <mat-expansion-panel *ngIf="getColorizerType() !== 'palette'" expanded="true">
+            <mat-expansion-panel *ngIf="paletteSelected === false" expanded="true">
                 <mat-expansion-panel-header>
                     <mat-panel-title fxLayoutAlign="space-between center" fxLayoutGap="0.5rem">
                         <mat-icon>looks</mat-icon>
--- a/projects/core/src/lib/layers/symbology/raster-symbology-editor/raster-symbology-editor.component.html
+++ b/projects/core/src/lib/layers/symbology/raster-symbology-editor/raster-symbology-editor.component.html
@@ -136,7 +136,6 @@
                     </mat-panel-title>
                 </mat-expansion-panel-header>
                 <ng-template matExpansionPanelContent>
-                    <!-- <geoengine-colorizer-editor [ngModel]="symbology.colorizer" (ngModelChange)="updateColorizer($event)"></geoengine-colorizer-editor> -->
                 </ng-template>
             </mat-expansion-panel>
         </mat-accordion>
--- a/projects/core/src/lib/layers/symbology/raster-symbology-editor/raster-symbology-editor.component.html
+++ b/projects/core/src/lib/layers/symbology/raster-symbology-editor/raster-symbology-editor.component.html
@@ -79,11 +79,10 @@
                             [multiple]="false"
                             [ngModel]="getColorizerType()"
                             (ngModelChange)="updateColorizerType($event)"
-                            [disabled]="getColorizerType() === 'palette'"
                         >
                             <mat-button-toggle fxFlexFill value="linearGradient">Linear Gradient</mat-button-toggle>
                             <mat-button-toggle fxFlexFill value="logarithmicGradient">Logarithmic Gradient</mat-button-toggle>
-                            <mat-button-toggle fxFlexFill value="palette" [disabled]="true">Palette</mat-button-toggle>
+                            <mat-button-toggle fxFlexFill value="palette">Palette</mat-button-toggle>
                         </mat-button-toggle-group>
                     </p>
                     <div
@@ -94,7 +93,7 @@
                 </ng-template>
             </mat-expansion-panel>
 
-            <mat-expansion-panel expanded="true">
+            <mat-expansion-panel *ngIf="getColorizerType() !== 'palette'" expanded="true">
                 <mat-expansion-panel-header>
                     <mat-panel-title fxLayoutAlign="space-between center" fxLayoutGap="0.5rem">
                         <mat-icon>looks</mat-icon>
@@ -113,7 +112,7 @@
                         <mat-progress-bar mode="indeterminate" *ngIf="histogramLoading | async"></mat-progress-bar>
                     </div>
                     <div class="actions">
-                        <button mat-raised-button class="button" color="primary" (click)="updateHistogram()">Update Histogram</button>
+                        <button mat-raised-button class="button" color="primary" (click)="updateHistogram()">Calculate Histogram</button>
                     </div>
                     <mat-divider></mat-divider>
 
@@ -144,12 +143,12 @@
     </div>
 
     <div class="actions">
-        <button mat-raised-button color="primary" class="button" (click)="applyChanges()" [disabled]="unappliedChanges === false">
-            Apply
-        </button>
-
         <button mat-raised-button color="accent" class="button" (click)="resetChanges(layer)" [disabled]="unappliedChanges === false">
             Reset
+        </button>
+
+        <button mat-raised-button color="primary" class="button" (click)="applyChanges()" [disabled]="unappliedChanges === false">
+            Apply
         </button>
     </div>
 </div>
--- a/projects/core/src/lib/layers/symbology/raster-symbology-editor/raster-symbology-editor.component.html
+++ b/projects/core/src/lib/layers/symbology/raster-symbology-editor/raster-symbology-editor.component.html
@@ -145,6 +145,7 @@
     </div>
 
     <div class="actions">
+        <span class="hint" *ngIf="unappliedChanges === true">Changes not reflected in map view!</span>
         <button mat-raised-button color="primary" id="applyButton" (click)="applyChanges()" [disabled]="unappliedChanges === false">
             Apply changes
         </button>
--- a/projects/core/src/lib/layers/symbology/raster-symbology-editor/raster-symbology-editor.component.html
+++ b/projects/core/src/lib/layers/symbology/raster-symbology-editor/raster-symbology-editor.component.html
@@ -1,133 +1,152 @@
 <geoengine-sidenav-header>Symbology Editor</geoengine-sidenav-header>
 
-<geoengine-dialog-help>
-    <p>The Symbology Editor enables customization of the style for raster layers.</p>
-    <p>
-        The <i>Global Layer Properties</i> define the default visualization parameters. The layer <i>Opacity</i> is adjustable in a range
-        from 0 to 100 %. You can choose a <i>NoData</i> color for pixels with the nodata value. Use the picker tool to select the desired
-        RGB color and opacity. This also applies to the <i>Overflow</i> color, which indicates the pixels with values without coloring
-        rules.
-    </p>
-    <p>
-        The <i>Color Map</i> section provides an overview of the pixel values with a frequency plot, which also allows to adapt the color
-        with respect to the raster values. The plot refers to the field of view shown in the map. If <i>Sync map and histogram </i> is
-        turned on, the histogram updates if the map view changes. To specify the value range of interest, you can set a minimum and maximum
-        pixel value. You can choose a color ramp from a variety of color schemes (<i>Colormap name</i>) and reverse it, if desired.
-        Additionally, different functions for the step distribution can be selected (linear, logarithmic, square root function, square
-        function). Consider that the logarithmic function requires positive values (>0). The number of Color steps is also kept flexible and
-        can be set to a number between 2 and 16. Click <i>Create color table</i> to apply your adjustments.
-    </p>
-    <p>
-        The <i> Color Table section </i> allows fine grained changes to colors. The gradient defines the interpolation between values. You
-        can dynamically add and remove color steps by clicking the minus and plus symbols or select distinct RGBA values for a specific
-        color step value.
-    </p>
-</geoengine-dialog-help>
-
-<mat-accordion multi="true">
-    <mat-expansion-panel expanded="true">
-        <mat-expansion-panel-header>
-            <mat-panel-title fxLayoutAlign="space-between center" fxLayoutGap="0.5rem">
-                <mat-icon>map</mat-icon>
-                <span fxFlex>Global Layer Properties</span>
-            </mat-panel-title>
-        </mat-expansion-panel-header>
-        <ng-template matExpansionPanelContent>
-            <div fxLayout="row" fxLayoutAlign="space-between center" fxLayoutGap="0.5rem">
-                <span>Opacity</span>
-                <mat-slider fxFlex #slo (change)="updateOpacity($event)" [value]="getOpacity()" max="100" min="1" step="1" thumbLabel>
-                </mat-slider>
-                <span>{{ slo.displayValue }} %</span>
-            </div>
-            <mat-divider></mat-divider>
-            <div>
-                <geoengine-color-attribute-input
-                    [ngModel]="getNoDataColor()"
-                    (ngModelChange)="updateNoDataColor($event)"
-                    attributePlaceholder=""
-                    [readonlyAttribute]="true"
-                >
-                </geoengine-color-attribute-input>
-                <geoengine-color-attribute-input
-                    [ngModel]="getDefaultColor()"
-                    (ngModelChange)="updateDefaultColor($event)"
-                    attributePlaceholder=""
-                    [readonlyAttribute]="true"
-                >
-                </geoengine-color-attribute-input>
-            </div>
-
-            <p fxLayout="row" fxLayoutAlign="space-between center" fxLayoutGap="0.5rem">
-                <span>Colorizer:</span>
-                <mat-button-toggle-group
-                    id="gradientType"
-                    fxFlex
-                    [multiple]="false"
-                    [ngModel]="getColorizerType()"
-                    (ngModelChange)="updateColorizerType($event)"
-                    [disabled]="getColorizerType() === 'palette'"
-                >
-                    <mat-button-toggle fxFlex value="linearGradient">Linear Gradient</mat-button-toggle>
-                    <mat-button-toggle fxFlex value="logarithmicGradient">Logarithmic Gradient</mat-button-toggle>
-                    <mat-button-toggle fxFlex value="palette" [disabled]="true">Palette</mat-button-toggle>
-                </mat-button-toggle-group>
+<div class="outer" fxLayout="column">
+    <div fxFlex class="container">
+        <geoengine-dialog-help>
+            <p>The Symbology Editor enables customization of the style for raster layers.</p>
+            <p>
+                The <i>Global Layer Properties</i> define the default visualization parameters. The layer <i>Opacity</i> is adjustable in a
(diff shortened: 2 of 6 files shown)


ClockworkOnion:
[[GEO BON] 4.1.1 Color palette switching (Raster)](https://trello.com/c/NJWQaDfb/1069-geo-bon-411-color-palette-switching-raster)

ChristianBeilschmidt (changes requested):

projects/core/src/lib/layers/symbology/raster-symbology-editor/raster-symbology-editor.component.html
```diff
@@ -71,6 +72,9 @@
                     <mat-button-toggle fxFlex value="palette" [disabled]="true">Palette</mat-button-toggle>
                 </mat-button-toggle-group>
             </p>
+            <button mat-raised-button color="primary" id="applyButton" (click)="applyChanges()" [disabled]="unappliedChanges === false">
+                Apply changes
+            </button>
```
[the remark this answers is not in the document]
ClockworkOnion: Added an always visible apply button mimicking the layout of Raster Vector Join component 389d86d8c0764b93d9a3e72e6c3f89d644ea15bd
The button applies changes from both Global Layer Properties as well as Update Color Map

projects/core/src/lib/layers/symbology/raster-symbology-editor/raster-symbology-editor.component.scss
```diff
@@ -24,3 +24,17 @@ mat-slider {
...
+
+#applyButton {
+    margin-right: 0rem;
+    margin-top: 1rem;
+    margin-bottom: 1rem;
+}
+
+.attention {
+    background-color: #ffeeee;
+}
```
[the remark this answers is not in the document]
ClockworkOnion: Removed the warning color and added a hint message in 51e5465ab293c014b6bfeb0621705973210b01ce but perhaps the message is unnecessary when the button is always in view?
image

projects/core/src/lib/colors/color-map-selector/color-map-selector.component.html
```diff
@@ -72,8 +72,8 @@
             </table>
         </div>
 
-        <mat-action-row>
+        <!-- <mat-action-row>
```
ChristianBeilschmidt, line 75: Please add a TODO here that this will be instantiated to populate the color table

projects/core/src/lib/layers/symbology/raster-symbology-editor/raster-symbology-editor.component.html
```diff
@@ -1,129 +1,153 @@
...
+                </ng-template>
+            </mat-expansion-panel>
+
+            <mat-expansion-panel expanded="true">
+                <mat-expansion-panel-header>
+                    <mat-panel-title fxLayoutAlign="space-between center" fxLayoutGap="0.5rem">
+                        <mat-icon>looks</mat-icon>
+                        <span fxFlex>Update Color Map</span>
+                    </mat-panel-title>
+                </mat-expansion-panel-header>
```
ChristianBeilschmidt, line 102: It would be great to only show the colormap creation for the gradients

projects/core/src/lib/layers/symbology/raster-symbology-editor/raster-symbology-editor.component.scss
```diff
@@ -24,3 +19,29 @@ mat-slider {
         margin: 0 auto;
     }
 }
+
+#gradientType {
+    flex-wrap: wrap;
```
ChristianBeilschmidt, line 24: grafik

1. This looks bad. It would be better to show the options as one item per row.
2. Please use the Angular flex layout selectors (https://github.com/angular/flex-layout) on the components when possible.

projects/core/src/lib/layers/symbology/raster-symbology-editor/raster-symbology-editor.component.html
```diff
@@ -1,129 +1,153 @@
...
-                <span fxFlex>Color Table</span>
-            </mat-panel-title>
-        </mat-expansion-panel-header>
-        <ng-template matExpansionPanelContent>
-            <!-- <geoengine-colorizer-editor [ngModel]="symbology.colorizer" (ngModelChange)="updateColorizer($event)"></geoengine-colorizer-editor> -->
-        </ng-template>
-    </mat-expansion-panel>
-</mat-accordion>
+    <div class="actions">
+        <span class="hint" *ngIf="unappliedChanges === true">Changes not reflected in map view!</span>
```
ChristianBeilschmidt, line 148:
```suggestion
        <span class="hint" *ngIf="unappliedChanges === true">Changes are not yet reflected in map view!</span>
```

ChristianBeilschmidt: But actually it is quite clear because you see the Apply changes button that you can click on all the time.

Maybe you can remove the message.

Moreover, maybe we should add a reset button there instead (not in primary color). What do you think?

projects/core/src/lib/layers/symbology/raster-symbology-editor/raster-symbology-editor.component.html
```diff
@@ -1,129 +1,153 @@
...
+                                [chartData]="histogramData"
+                                (interactionChange)="updateBounds($any($event))"
+                            ></geoengine-vega-viewer>
+                        </ng-container>
+                        <mat-progress-bar mode="indeterminate" *ngIf="histogramLoading | async"></mat-progress-bar>
+                    </div>
+                    <div>
+                        <mat-slide-toggle [checked]="histogramAutoReload" (change)="histogramAutoReload = $event.checked">
+                            Sync map and histogram
+                        </mat-slide-toggle>
```
ChristianBeilschmidt, line 117: You still need to change this to only compute the histogram when you click a "Calculate histogram" button.

ChristianBeilschmidt (changes requested):

projects/core/src/lib/layers/symbology/raster-symbology-editor/raster-symbology-editor.component.html
ClockworkOnion: I added a reset button that reloads the component. It's using the "accent" color for now.
Also removed the message about unapplied changes. f2c0f44af28f79d4d5bb91812260706d2c463892
image

projects/core/src/lib/layers/symbology/raster-symbology-editor/raster-symbology-editor.component.html
ClockworkOnion: Changed it to a button now f2c0f44af28f79d4d5bb91812260706d2c463892

projects/core/src/lib/layers/symbology/raster-symbology-editor/raster-symbology-editor.component.scss
ClockworkOnion: Each button is now displayed in its own row, and I moved the "Colorizer" text to the top as well. Also added a divider between the color pickers and colorizer settings. f2c0f44af28f79d4d5bb91812260706d2c463892
image

projects/core/src/lib/colors/color-map-selector/color-map-selector.component.html
ClockworkOnion: Replaced the commented out section with a TODO f2c0f44af28f79d4d5bb91812260706d2c463892 (Did I understand this right?)

projects/core/src/lib/layers/symbology/raster-symbology-editor/raster-symbology-editor.component.html
ClockworkOnion: If I am understanding this right, the bottom component "Update Color Map" should be hidden when 'Palette' is selected instead of 'Linear Gradient' or 'Logarithmic Gradient'.
I think this will be easy to achieve with an *ngIf, once Palette mode is implemented. For now, since the `getColorizerType()` method depends on the symbology of the layer being an instance of `PaletteColorizer` (which makes sense of course), I don't think there is a way of returning 'palette' from `getColorizerType()` without an intrusive workaround that would ultimately be reversed later.
So my suggestion would be to wait with this change until the next step, where I'll try to implement the user selected palette

projects/core/src/lib/layers/symbology/raster-symbology-editor/raster-symbology-editor.component.html
ChristianBeilschmidt: The important thing is, that it is *not* calculated automatically upfront.
The user has to calculate it actively so that the user can decide whether to do it or not.
This is why I meant "Calculate histogram" since you cannot update it upfront.

projects/core/src/lib/layers/symbology/raster-symbology-editor/raster-symbology-editor.component.html
ChristianBeilschmidt: Please switch the places (left to right) – it sees more natural, doesn't it?

Moreover, clicking reset (good idea by the way) should not recompute the histogram.

projects/core/src/lib/layers/symbology/raster-symbology-editor/raster-symbology-editor.component.html
ChristianBeilschmidt: `getColorizerType()` seems to already return `palette` as one option. Otherwise, the button wouldn't be selected.

projects/core/src/lib/layers/symbology/raster-symbology-editor/raster-symbology-editor.component.html
```diff
@@ -1,129 +1,155 @@
...
+            <!-- TODO: implement palette editor -->
+            <mat-expansion-panel expanded="false" *ngIf="symbology.colorizer.isDiscrete()">
+                <mat-expansion-panel-header>
+                    <mat-panel-title fxLayoutAlign="space-between center" fxLayoutGap="0.5rem">
+                        <mat-icon>color_lens</mat-icon>
+                        <span fxFlex>Color Table</span>
+                    </mat-panel-title>
+                </mat-expansion-panel-header>
+                <ng-template matExpansionPanelContent>
+                    <!-- <geoengine-colorizer-editor [ngModel]="symbology.colorizer" (ngModelChange)="updateColorizer($event)"></geoengine-colorizer-editor> -->
```
ChristianBeilschmidt, line 140:
```suggestion
```

ChristianBeilschmidt (changes requested):

projects/core/src/lib/colors/color-map-selector/color-map-selector.component.html
ClockworkOnion: After all. :) I put the commented out method back in for future reference df5f7fdfc5f3388cdcb12b1a05092ac86301c885

projects/core/src/lib/layers/symbology/raster-symbology-editor/raster-symbology-editor.component.html
ClockworkOnion: Switched the button positions, and the histogram isn't calculated on a reset df5f7fdfc5f3388cdcb12b1a05092ac86301c885

projects/core/src/lib/layers/symbology/raster-symbology-editor/raster-symbology-editor.component.html
ClockworkOnion: I removed the automatic calculation and changed the button text to "Calculate Histogram" df5f7fdfc5f3388cdcb12b1a05092ac86301c885

projects/core/src/lib/layers/symbology/raster-symbology-editor/raster-symbology-editor.component.html
ClockworkOnion: In 29ff0dfe42bb1d8e1d1f657b4dfe5934e991ea45 the panel gets hidden when palette is selected

ChristianBeilschmidt: grafik

Unfortunately, it is not.

projects/core/src/lib/layers/symbology/raster-symbology-editor/raster-symbology-editor.component.ts
```diff
@@ -307,6 +343,14 @@ export class RasterSymbologyEditorComponent implements OnChanges, OnDestroy, Aft
         this.updateLayerMaxValue(breakpoints[breakpoints.length - 1].value);
     }
```
ChristianBeilschmidt, line 345: grafik

The empty space and the button somehow looks bad or broken, because it looks like we just randomly put a button there.

Can we change that somehow that it looks good with and without a histogram plot being present?

Maybe it is sufficient to align the button to the right or left or reduce the empty space ontop… I'm not sure. But we shouldn't leave it as it because the rest looks good now.

ChristianBeilschmidt (changes requested):

projects/core/src/lib/layers/symbology/raster-symbology-editor/raster-symbology-editor.component.html
ClockworkOnion: Fixed the bug where it's not hidden when Palette was selected as default [3f22461](https://github.com/geo-engine/geoengine-ui/pull/513/commits/3f224612d1c12d14216ac47e94902993d7b3c84d)

projects/core/src/lib/layers/symbology/raster-symbology-editor/raster-symbology-editor.component.ts
ClockworkOnion: Changed it so at first there is no empty space, the button is left aligned and says "Create Histogram". When the first histogram is created, the button is then right aligned below the histogram and says "Update Histogram" as before
image

[c618e4d](https://github.com/geo-engine/geoengine-ui/pull/513/commits/c618e4d5884fe8dc8d72f3f266b4212c84d8c86e)

ChristianBeilschmidt: diff --git a/projects/core/src/lib/layers/symbology/raster-symbology-editor/raster-symbology-editor.component.html b/projects/core/src/lib/layers/symbology/raster-symbology-editor/raster-symbology-editor.component.html
index 5f5cbd45..fb7f7e97 100644
--- a/projects/core/src/lib/layers/symbology/raster-symbology-editor/raster-symbology-editor.component.html
+++ b/projects/core/src/lib/layers/symbology/raster-symbology-editor/raster-symbology-editor.component.html
@@ -111,17 +111,16 @@
                         </ng-container>
                         <mat-progress-bar mode="indeterminate" *ngIf="histogramLoading | async"></mat-progress-bar>
                     </div>
-                    <div class="actions" *ngIf="histogramCreated; then createdBlock; else notCreatedBlock"></div>
-                    
-                    <ng-template #createdBlock>
-                        <div class="actions">
-                            <button mat-raised-button class="button" color="primary" (click)="updateHistogram()">Update Histogram</button>
-                        </div>
-                    </ng-template>
-                    
-                    <ng-template #notCreatedBlock>
-                        <button mat-raised-button class="button" color="primary" (click)="updateHistogram()">Create Histogram</button>
-                    </ng-template>
+                    <div class="histogram-actions">
+                        <button mat-stroked-button class="button" (click)="updateHistogram()">
+                            <ng-container *ngIf="histogramCreated; then createdBlock; else notCreatedBlock"></ng-container>
+                            
+                            <ng-template #createdBlock>Update Histogram</ng-template>
+                            
+                            <ng-template #notCreatedBlock>Create Histogram</ng-template>
+                        </button>
+                    </div>
+
                     <mat-divider></mat-divider>
 
                     <geoengine-color-map-selector
@@ -149,9 +148,7 @@
     </div>
 
     <div class="actions">
-        <button mat-raised-button color="accent" class="button" (click)="resetChanges(layer)" [disabled]="unappliedChanges === false">
-            Reset
-        </button>
+        <button mat-raised-button class="button" (click)="resetChanges(layer)" [disabled]="unappliedChanges === false">Reset</button>
 
         <button mat-raised-button color="primary" class="button" (click)="applyChanges()" [disabled]="unappliedChanges === false">
             Apply
diff --git a/projects/core/src/lib/layers/symbology/raster-symbology-editor/raster-symbology-editor.component.scss b/projects/core/src/lib/layers/symbology/raster-symbology-editor/raster-symbology-editor.component.scss
index dd88bd7a..156b39bf 100644
--- a/projects/core/src/lib/layers/symbology/raster-symbology-editor/raster-symbology-editor.component.scss
+++ b/projects/core/src/lib/layers/symbology/raster-symbology-editor/raster-symbology-editor.component.scss
@@ -20,6 +20,10 @@ mat-slider {
     }
 }
 
+.histogram-actions {
+    text-align: center;
+}
+
 #gradientType {
     flex-wrap: wrap;
 }

ChristianBeilschmidt: This still looks like the button was placed by accident.

What about this?

```diff
diff --git a/projects/core/src/lib/layers/symbology/raster-symbology-editor/raster-symbology-editor.component.html b/projects/core/src/lib/layers/symbology/raster-symbology-editor/raster-symbology-editor.component.html
index 5f5cbd45..fb7f7e97 100644
--- a/projects/core/src/lib/layers/symbology/raster-symbology-editor/raster-symbology-editor.component.html
+++ b/projects/core/src/lib/layers/symbology/raster-symbology-editor/raster-symbology-editor.component.html
@@ -111,17 +111,16 @@
                         </ng-container>
                         <mat-progress-bar mode="indeterminate" *ngIf="histogramLoading | async"></mat-progress-bar>
                     </div>
-                    <div class="actions" *ngIf="histogramCreated; then createdBlock; else notCreatedBlock"></div>
-                    <!-- Only displayed before any histogram is created -->
-                    <ng-template #createdBlock>
-                        <div class="actions">
-                            <button mat-raised-button class="button" color="primary" (click)="updateHistogram()">Update Histogram</button>
-                        </div>
-                    </ng-template>
-                    <!-- Only displayed after histogram is created -->
-                    <ng-template #notCreatedBlock>
-                        <button mat-raised-button class="button" color="primary" (click)="updateHistogram()">Create Histogram</button>
-                    </ng-template>
+                    <div class="histogram-actions">
+                        <button mat-stroked-button class="button" (click)="updateHistogram()">
+                            <ng-container *ngIf="histogramCreated; then createdBlock; else notCreatedBlock"></ng-container>
+                            <!-- Only displayed before any histogram is created -->
+                            <ng-template #createdBlock>Update Histogram</ng-template>
+                            <!-- Only displayed after histogram is created -->
+                            <ng-template #notCreatedBlock>Create Histogram</ng-template>
+                        </button>
+                    </div>
+
                     <mat-divider></mat-divider>
 
                     <geoengine-color-map-selector
@@ -149,9 +148,7 @@
     </div>
 
     <div class="actions">
-        <button mat-raised-button color="accent" class="button" (click)="resetChanges(layer)" [disabled]="unappliedChanges === false">
-            Reset
-        </button>
+        <button mat-raised-button class="button" (click)="resetChanges(layer)" [disabled]="unappliedChanges === false">Reset</button>
 
         <button mat-raised-button color="primary" class="button" (click)="applyChanges()" [disabled]="unappliedChanges === false">
             Apply
diff --git a/projects/core/src/lib/layers/symbology/raster-symbology-editor/raster-symbology-editor.component.scss b/projects/core/src/lib/layers/symbology/raster-symbology-editor/raster-symbology-editor.component.scss
index dd88bd7a..156b39bf 100644
--- a/projects/core/src/lib/layers/symbology/raster-symbology-editor/raster-symbology-editor.component.scss
+++ b/projects/core/src/lib/layers/symbology/raster-symbology-editor/raster-symbology-editor.component.scss
@@ -20,6 +20,10 @@ mat-slider {
     }
 }
 
+.histogram-actions {
+    text-align: center;
+}
+
 #gradientType {
     flex-wrap: wrap;
 }

```

Cf. `git apply` for applying this.

ClockworkOnion: Ok, I applied these changes in [31041bf](https://github.com/geo-engine/geoengine-ui/pull/513/commits/31041bf8a9dd6d3f27a9956618b1e6fb56c9cebb)

projects/core/src/lib/layers/symbology/raster-symbology-editor/raster-symbology-editor.component.ts
```diff
@@ -162,9 +183,23 @@ export class RasterSymbologyEditorComponent implements OnChanges, OnDestroy, Aft
...
+        this.colorMapSelector.applyChanges();
+        this.unappliedChanges = false;
         this.update();
     }
 
+    getNotified(): void {
+        this.unappliedChanges = true;
+    }
+
+    resetChanges(layer: Layer): void {
```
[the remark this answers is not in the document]
ClockworkOnion: I tried at first to implement it like that and preserve the histogram. But I'm not sure if the last emitted value from the histogram observable is still available somewhere. If it is, I couldn't find how/where to access it. If it's no longer accessible I would need to store it somewhere outside of the component each time a new histogram is created... but where would be a good place?

Another way to go about resetting the changes while preserving the histogram might be to store the values/settings for all the other sliders beforehand, and make the reset button fill the previous values back in. That way the component wouldn't be recreated and the histogram isn't touched. But perhaps it could be a bit intrusive to the rest of the code if done like this.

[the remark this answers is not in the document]
ChristianBeilschmidt:
```diff
diff --git a/projects/core/src/lib/layers/symbology/raster-symbology-editor/raster-symbology-editor.component.ts b/projects/core/src/lib/layers/symbology/raster-symbology-editor/raster-symbology-editor.component.ts
index 745c495f..37b4927c 100644
--- a/projects/core/src/lib/layers/symbology/raster-symbology-editor/raster-symbology-editor.component.ts
+++ b/projects/core/src/lib/layers/symbology/raster-symbology-editor/raster-symbology-editor.component.ts
@@ -197,7 +197,10 @@ export class RasterSymbologyEditorComponent implements OnChanges, OnDestroy, Aft
     }
 
     resetChanges(layer: Layer): void {
-        this.layoutService.setSidenavContentComponent({component: RasterSymbologyEditorComponent, config: {layer}});
+        this.layoutService.setSidenavContentComponent({
+            component: RasterSymbologyEditorComponent,
+            config: {layer, histogramData: this.histogramData, histogramCreated: this.histogramCreated},
+        });
     }
 
     getNoDataColor(): ColorAttributeInput {
```

Please check if there are no undesired side effects by doing this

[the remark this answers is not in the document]
ClockworkOnion: Thanks! That's a very nice way of keeping the histogram, didn't know this was possible. I tested the functionality and found no side effects.

ChristianBeilschmidt:
bors merge

PR #541, opened 2022-12-12
replace baskets with gfbio collections provider

michaelmattig:


michaelmattig:
[NFDI4BIO-200 Implement Connection to Data Collection Service](https://trello.com/c/UTTy9IM6/1136-nfdi4bio-200-implement-connection-to-data-collection-service)

projects/core/src/lib/layer-collections/layer-collection.service.ts
```diff
@@ -45,4 +76,186 @@ export class LayerCollectionService {
...
+    /**
+     * Add all layers (directly) contained in a layer collection to the current project.
+     */
+    addCollectionLayersToProject(collection: LayerCollectionDict): void {
+        const layersObservable = collection.items
+            .filter((layer) => layer.type === 'layer')
+            .map((layer) => layer as LayerCollectionLayerDict)
+            .map((layer) => this.resolveLayer(layer.id));
+
+        // TODO: lookup in parallel
```
ChristianBeilschmidt, line 89: mergeMap?

projects/core/src/lib/layer-collections/layer-collection.service.ts
```diff
@@ -45,4 +76,186 @@ export class LayerCollectionService {
...
+        zip(layersObservable)
+            .pipe(map((layers) => this.projectService.addLayers(layers)))
+            .subscribe();
+    }
+
+    /**
+     * Add a layer to the current project.
+     */
+    addLayerToProject(layerId: ProviderLayerIdDict): void {
```
ChristianBeilschmidt, line 98: Maybe use `subscribeAndProvide` to make it optional to await on it?

We should move the method from he ProjectService to a general util then, I guess.

projects/core/src/lib/layer-collections/layer-collection.service.ts
```diff
@@ -45,4 +76,186 @@ export class LayerCollectionService {
             .getSessionStream()
```
ChristianBeilschmidt, line 76: The session stream emits on session changes? Do you want `getSessionForRequest`, which only emits once?

ChristianBeilschmidt: Maybe look at all the other `getSessionStream()` in this file as well if this is intended.

projects/core/src/lib/layer-collections/layer-collection.service.ts
```diff
@@ -45,4 +76,186 @@ export class LayerCollectionService {
...
+                    });
+                } else {
+                    // TODO: implement plots, etc.
+                    throw new Error('Adding this workflow type is unimplemented, yet');
+                }
+            }),
+        );
+    }
+
+    private doAddLayerToProject(layer: LayerDict, workflowId: UUID, resultDescriptorDict: ResultDescriptorDict): void {
```
ChristianBeilschmidt, line 154: When is `addLayerToProject` and `doAddLayerToProject` used?

projects/core/src/lib/layer-collections/layer-collection.service.ts
```diff
@@ -45,4 +76,186 @@ export class LayerCollectionService {
...
+                    type: 'linearGradient',
+                    breakpoints: [
+                        {value: 1, color: [0, 0, 0, 255]},
+                        {value: 255, color: [255, 255, 255, 255]},
+                    ],
+                    defaultColor: [0, 0, 0, 0],
+                    noDataColor: [0, 0, 0, 0],
+                },
+            };
+        }
```
ChristianBeilschmidt, line 249: Is there a chance to generate a palette for a classification?

projects/core/src/lib/layer-collections/layer-collection.service.ts
```diff
@@ -45,4 +76,186 @@ export class LayerCollectionService {
...
             .pipe(mergeMap((session) => this.backend.getWorkflowMetadata(workflowId, session.sessionToken)));
     }
+
+    /**
+     * Add all layers (directly) contained in a layer collection to the current project.
+     */
+    addCollectionLayersToProject(collection: LayerCollectionDict): void {
+        const layersObservable = collection.items
+            .filter((layer) => layer.type === 'layer')
+            .map((layer) => layer as LayerCollectionLayerDict)
```
ChristianBeilschmidt, line 86: Do you get the "wrong" type here? You basically just typecast-map, right?

projects/dashboards/gfbio/src/app/app.component.ts
```diff
@@ -279,11 +279,16 @@ export class AppComponent implements OnInit, AfterViewInit {
...
-            const basketId = p.get('basket_id');
-            if (basketId != null) {
-                this.basketService.handleBasket(basketId).subscribe((result) => {
-                    this.dialog.open(BasketDialogComponent, {data: {result}});
-                });
+            const collectionId = p.get('collectionId');
+            if (collectionId != null) {
+                // this.basketService.handleBasket(collectionId).subscribe((result) => {
+                //     this.dialog.open(BasketDialogComponent, {data: {result}});
+                // });
```
ChristianBeilschmidt, line 286:
```suggestion
```

projects/dashboards/gfbio/src/app/app.component.ts
```diff
@@ -79,6 +78,8 @@ export class AppComponent implements OnInit, AfterViewInit {
 
     private windowHeight$ = new BehaviorSubject<number>(window.innerHeight);
 
+    private GFBIO_COLLECTIONS_DATA_PROVIDER_ID = 'f64e2d5b-3b80-476a-83f5-c330956b2909';
```
ChristianBeilschmidt, line 81: I guess it is not a task for now, but we should think about configuring our GIS.

projects/dashboards/gfbio/src/app/gfbio-collection/gfbio-collection-dialog.component.html
```diff
@@ -1,26 +1,24 @@
 <geoengine-dialog-header>VAT - The Visualization, Analysis and Transformation System</geoengine-dialog-header>
 <mat-dialog-content>
-    <h2>Basket Import</h2>
-    <p>Here is a summary of the imported datasets from your basket!</p>
+    <h2>GFBio Collection Import</h2>
+    <p>Here is a summary of the imported datasets from your collection!</p>
```
ChristianBeilschmidt, line 4:
```suggestion
    <p>Here is a summary of the imported datasets from your collection:</p>
```

projects/dashboards/gfbio/src/app/gfbio-collection/gfbio-collection-dialog.component.ts
```diff
@@ -0,0 +1,75 @@
...
+
+    layerStatus(item: LayerCollectionItemDict): LayerStatus {
+        const status = item.properties.find(([a, _]) => a === 'status');
+
+        if (status) {
+            return this.statusFromString(status[1]);
+        } else {
+            return LayerStatus.Error;
+        }
+    }
```
ChristianBeilschmidt, line 40: Is the status specific to GFBioCollectionItems? If not, maybe move it to core?

projects/dashboards/gfbio/src/app/gfbio-collection/gfbio-collection-dialog.component.ts
```diff
@@ -0,0 +1,75 @@
...
+
+    private statusFromString(status: string): LayerStatus {
+        switch (status) {
+            case 'ok':
+                return LayerStatus.Ok;
+            case 'unavailable':
+                return LayerStatus.Unavailable;
+            default:
+                return LayerStatus.Error;
```
ChristianBeilschmidt, line 72: Is it wise to log an error if the status is not error, but we fall back to it?
Falling back to an error as it is now makes sense, since there would be something wrong anyway.

projects/core/src/lib/backend/backend.model.ts
```diff
@@ -548,6 +548,7 @@ export interface LayerCollectionItemDict {
     id: ProviderLayerIdDict | ProviderLayerCollectionIdDict;
     name: string;
     description: string;
+    properties: [string, string][];
```
ChristianBeilschmidt, line 551:
```suggestion
    properties: Array<[string, string]>;
```

would be my preference.
Where is Clippy when we would need it :smile: .

ChristianBeilschmidt (changes requested):

projects/dashboards/gfbio/src/app/gfbio-collection/gfbio-collection-dialog.component.html
```diff
@@ -1,26 +1,24 @@
...
                 <mat-header-cell *matHeaderCellDef>Status </mat-header-cell>
                 <mat-cell *matCellDef="let element"
-                    ><span class="{{ element.status }}">{{ element.status }}</span></mat-cell
+                    ><span class="{{ layerStatus(element) }}">{{ layerStatus(element) }}</span></mat-cell
                 >
             </ng-container>
 
-            <mat-header-row *matHeaderRowDef="['title', 'status']"></mat-header-row>
-            <mat-row *matRowDef="let row; columns: ['title', 'status']"></mat-row>
+            <mat-header-row *matHeaderRowDef="['name', 'status']"></mat-header-row>
```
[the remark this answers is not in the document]
michaelmattig: it doesn't really add any functionality because users can just remove the layer once it was added. If we have time/capacity we can do it after everything else works. Otherwise we can do it in a future project.

projects/core/src/lib/layer-collections/layer-collection.service.ts
michaelmattig: this is a private helper method, used by addLayerToProject which is public

projects/dashboards/gfbio/src/app/app.component.ts
michaelmattig: Once we have stable names for stuff, we should use the stable name of the GFBio Collections provider here instead of a cryptic UUID. IMHO this here is nothing that needs configuration.

projects/core/src/lib/layer-collections/layer-collection.service.ts
michaelmattig: a collection can also contain child collections. This method only adds the directly contained layers, that's why I filter here.

projects/core/src/lib/layer-collections/layer-collection-list/layer-collection-list.component.ts
```diff
@@ -136,31 +119,7 @@ export class LayerCollectionListComponent implements OnInit, AfterViewInit {
...
-                        this.addRasterLayer(layer.name, workflowId, resultDescriptorDict as RasterResultDescriptorDict, layer.symbology);
-                    } else {
-                        // TODO: implement plots, etc.
-                        this.notificationService.error('Adding this workflow type is unimplemented, yet');
-                    }
-                },
-                error: (requestError) => this.handleError(requestError.error, layerId.layerId),
-            });
+        this.layerService.addLayerToProject(layerId);
     }
```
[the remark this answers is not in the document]
michaelmattig: I added a loading indicator 8379c66470c4beb3fdc9118ef3a17eba367299f8

projects/core/src/lib/layer-collections/layer-collection-list/layer-collection-list.component.ts
```diff
@@ -169,111 +128,6 @@ export class LayerCollectionListComponent implements OnInit, AfterViewInit {
...
-    private handleError(error: GeoEngineErrorDict, workflowId: UUID): void {
-        let errorMessage = `No workflow found for id: ${workflowId}`;
-
-        if (error.error !== 'NoWorkflowForGivenId') {
-            errorMessage = `Unknown error -> ${error.error}: ${error.message}`;
-        }
-
-        this.notificationService.error(errorMessage);
-    }
 }
```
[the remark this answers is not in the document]
michaelmattig: I added a loading indicator 8379c66470c4beb3fdc9118ef3a17eba367299f8

projects/dashboards/gfbio/src/app/gfbio-collection/gfbio-collection-dialog.component.ts
```diff
@@ -0,0 +1,75 @@
...
+        this.dialogRef.close();
+    }
+
+    replaceLayers(): void {
+        const filteredCollection = this.collection;
+        this.collection.items = this.getOkLayers();
+
+        this.projectService
+            .clearLayers()
+            .pipe(mergeMap(() => of(this.layerService.addCollectionLayersToProject(filteredCollection))))
```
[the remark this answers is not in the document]
michaelmattig: I added a loading indicator 8379c66470c4beb3fdc9118ef3a17eba367299f8

projects/core/src/lib/layer-collections/layer-collection.service.ts
michaelmattig: my idea was to add the layers as soon as the metadata is retrieved instead of waiting for all to be finished, but I don't know if that really makes any difference.

projects/core/src/lib/layer-collections/layer-collection.service.ts
michaelmattig: 8379c66470c4beb3fdc9118ef3a17eba367299f8

projects/core/src/lib/layer-collections/layer-collection.service.ts
michaelmattig: We could generate one, if the measurement is available and is a classification. But to be fair, the default gradient also doesn't really make sense with a range from 1 to 255...

projects/dashboards/gfbio/src/app/gfbio-collection/gfbio-collection-dialog.component.ts
michaelmattig: What does it achieve to log this? We could introduce a fourth type "Status N/A", then users could report this as a bug. Because the only way that the status is missing in the layer properties is if there is a bug in the data provider.

projects/dashboards/gfbio/src/app/gfbio-collection/gfbio-collection-dialog.component.html
michaelmattig: 65dbbac9aaf94ea02c8d34d5acc86522d619ebd1

projects/dashboards/gfbio/src/app/app.component.ts
michaelmattig: 65dbbac9aaf94ea02c8d34d5acc86522d619ebd1

projects/core/src/lib/backend/backend.model.ts
michaelmattig: 65dbbac9aaf94ea02c8d34d5acc86522d619ebd1

projects/core/src/lib/layer-collections/layer-collection.service.ts
ChristianBeilschmidt: Currently, there is the danger to have a race condition when calling addLayer concurrently I guess. At least we would have to look into this.

With mergeMap, we would have at least all called concurrently, while for zip they are called in order or am I mistaking?

projects/dashboards/gfbio/src/app/gfbio-collection/gfbio-collection-dialog.component.html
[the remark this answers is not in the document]
ChristianBeilschmidt: Ok, let's keep this open and decide upon everything else is resolved.

projects/core/src/lib/layer-collections/layer-collection.service.ts
ChristianBeilschmidt: This seems right, but this also seems like a bug. Maybe you can take a look.

projects/core/src/lib/layer-collections/layer-collection.service.ts
ChristianBeilschmidt: But why do both exist here? It is not used by addLayerToProject.

It looks like two methods do the same thing slightly different.

projects/core/src/lib/layer-collections/layer-collection.service.ts
ChristianBeilschmidt: I don't mean the filter, but the `map`.

projects/dashboards/gfbio/src/app/gfbio-collection/gfbio-collection-dialog.component.ts
ChristianBeilschmidt: The intermediate way would be to log/output it that this should be reported. Then, we don't need another handling than `Error`. It will be visible as a notification and in the log menu.

michaelmattig: Do you mean notification with our NotificationService? 
What if there are multiple elements with error? One notification per error?
To whom should it be reported? Just like "Contact the administrator"?

projects/core/src/lib/layer-collections/layer-collection.service.ts
michaelmattig: you are right they were redundant e3d9720bde4caf0444077f523258e65bd9b984dc

projects/core/src/lib/layer-collections/layer-collection.service.ts
michaelmattig: Where do you see a race condition? I thought JavaScript only uses a single thread anyway.

> With mergeMap, we would have at least all called concurrently, while for zip they are called in order or am I mistaking?

no, zip only emits after all observables have emitted. But: we could start adding the first layer as soon as it (or rather: its metdata) is ready. I wanted to use `forkJoin` for this but couldn't get it to work, so I left a TODO here.

projects/dashboards/gfbio/src/app/gfbio-collection/gfbio-collection-dialog.component.ts
ChristianBeilschmidt: Yeah, I guess. Just plain and simple report what happened.

michaelmattig: 68aaaa107ab1538aa785539828ccfccf457074de

you can only see one error as toast, but in the log you can see the rest

ChristianBeilschmidt (approved):

ChristianBeilschmidt:
bors merge

PR #581, opened 2023-02-03
abort query when layer was deleted

michaelmattig:


michaelmattig:
[Workflow Abbruch bei Löschen eines Layers](https://trello.com/c/tFffLgDp/1127-workflow-abbruch-bei-l%C3%B6schen-eines-layers)

projects/core/src/lib/project/project.service.ts
```diff
@@ -874,24 +874,27 @@ export class ProjectService {
...
-    createQueryAbortStream(tileZoomLevel: number, tileExtent: Extent): Observable<void> {
+    createQueryAbortStream(layerId: number, tileZoomLevel: number, tileExtent: Extent): Observable<void> {
         const tileResolution = this.mapService.getView().getResolutionForZoom(tileZoomLevel);
 
         const observables: Array<Observable<any>> = [
             this.getTimeStream(),
             this.mapService.getViewportSizeStream(),
             this.userService.getSessionTokenForRequest(),
             this.getSpatialReferenceStream(),
+            this.getLayerStream(),
```
ChristianBeilschmidt, line 888: Woudln't it be enough to look at only one layer in `layers`?

ChristianBeilschmidt (changes requested):

projects/core/src/lib/project/project.service.ts
michaelmattig: is there a way to monitor a single layer? If there is, what is it?

ChristianBeilschmidt: Yeah, the entries/oversables in the `layers` variable. It is set in `layerChangesStream` (or similar) and completed in `removeLayerSubscriptions`.

michaelmattig: ok, but what is your idea here? I would need an observable for each layer. Like `layerDeletion$: Map<number, Subscription>()` and subscribe to it. Is it really worth to introduce this overhead, instead of just scanning the existing layers? it's not like we are removing hundreds of layers all the time.

If that's not what you meant, please give me a more detailed description of your desired solution.

michaelmattig: 5b0f8bf4f0801cb28cc37b4a8dc339c3eea94bb5

projects/core/src/lib/project/project.service.ts
```diff
@@ -874,24 +874,36 @@ export class ProjectService {
...
 
+        // create an observable that emits when the layer is removed
+        const layerStream = this.layers.get(layerId);
+        if (!layerStream) {
+            throw Error(`No layer stream found for layer id ${layerId}`);
+        }
+        const layerRemovedSubject = new ReplaySubject<boolean>();
+        layerStream.subscribe({
+            next: () => layerRemovedSubject.next(false),
+            complete: () => layerRemovedSubject.next(true),
```
ChristianBeilschmidt, line 891:
```suggestion
            complete: () => {
                layerRemovedSubject.next(true);
                layerRemovedSubject.complete();
            },
```
:see_no_evil:

ChristianBeilschmidt (changes requested):

ChristianBeilschmidt (approved):

ChristianBeilschmidt:
bors merge

PR #599, opened 2023-02-27
fix layer listing

ChristianBeilschmidt:


ChristianBeilschmidt:
[Fehlerhafte Darstellung in Layerliste bei langen Namen/Beschreibungen](https://trello.com/c/7zG9Pn0R/1148-fehlerhafte-darstellung-in-layerliste-bei-langen-namen-beschreibungen)

projects/core/src/lib/layer-collections/layer-collection-layer/layer-collection-layer.component.scss
```diff
@@ -4,17 +4,18 @@
 }
```
jdroenner, line 5: The details will get a scroll bar if the sidenav gets to small. I think this is fine. However there is a bit missing on the right...

image

maybe you could look into that?

jdroenner (changes requested):

projects/core/src/lib/layer-collections/layer-collection-layer/layer-collection-layer.component.scss
ChristianBeilschmidt: [be126a1](https://github.com/geo-engine/geoengine-ui/pull/599/commits/be126a16043a28e1a8b742ac2b895fd00fef40f4)

Scrolling seems to be strange, so instead, I let the texts break.

jdroenner (approved):

jdroenner:
bors merge

PR #621, opened 2023-04-27
fix oidc login when anonymous access is disabled

michaelmattig:


michaelmattig:
[KeyCloak aufsetzen und konfigurieren](https://trello.com/c/Rtx1wgbi/1285-keycloak-aufsetzen-und-konfigurieren)

projects/core/src/lib/users/oidc/oidc.component.ts
```diff
@@ -19,7 +20,7 @@ export class OidcComponent implements OnInit, OnDestroy {
 
     private pendingLoginRequest = false;
 
-    constructor(private readonly userService: UserService) {}
+    constructor(private readonly userService: UserService, private router: Router) {}
```
ChristianBeilschmidt, line 23:
```suggestion
    constructor(private readonly userService: UserService, private readonly router: Router) {}
```

projects/core/src/lib/users/oidc/oidc.component.ts
```diff
@@ -65,7 +66,10 @@ export class OidcComponent implements OnInit, OnDestroy {
                         this.user = session.user;
                     }
                 },
-                error: () => {},
+                error: () => {
+                    // guest login failed -> navigate to sign in
+                    this.router.navigate(['/signin']);
```
ChristianBeilschmidt, line 71: Does this work in any case? Maybe route to `/` such that this works in any app?

projects/core/src/lib/users/user.service.ts
```diff
@@ -325,6 +330,20 @@ export class UserService {
...
+
+    private checkOidcParametersAndLogin(): boolean {
+        const params = new URLSearchParams(window.location.search);
+        const sessionState = params.get('session_state');
+        const code = params.get('code');
+        const state = params.get('state');
+
+        if (sessionState && code && state) {
+            this.oidcLogin({sessionState, code, state}).pipe(first()).subscribe();
+            return true;
```
ChristianBeilschmidt, line 342: This looks… wrong.

Shouldn't you return an `Observable<boolean>` instead?

projects/core/src/lib/users/user.service.ts
```diff
@@ -49,19 +49,24 @@ export class UserService {
...
-                    error: (error) => {
-                        // only show error if we did not expect it
-                        if (error.error.error !== 'AnonymousAccessDisabled') {
-                            this.notificationService.error(error.error.message);
-                        }
-                        this.session$.next(undefined);
-                    },
-                });
+
+                const oidcLogin = this.checkOidcParametersAndLogin();
```
ChristianBeilschmidt, line 53: Don't you need to await something?

projects/gis/src/app/login/login.component.ts
```diff
@@ -50,17 +53,27 @@ export class LoginComponent implements OnInit, AfterViewInit, OnDestroy {
...
-                } else {
-                    this.user = session.user;
-                    this.formStatus$.next(FormStatus.LoggedIn);
-                }
-            });
+        // check if OIDC login is enabled
+        this.userService.oidcInit().subscribe(
+            (idr) => {
+                this.oidcUrl = idr.url;
+                this.formStatus$.next(FormStatus.Oidc);
```
ChristianBeilschmidt, line 60: Why don't we need the subscription here?

projects/gis/src/app/main/main.component.ts
```diff
@@ -248,15 +246,4 @@ export class MainComponent implements OnInit, AfterViewInit {
...
-    private handleQueryParameters(): void {
-        const params = new URLSearchParams(window.location.search);
-        const sessionState = params.get('session_state');
-        const code = params.get('code');
-        const state = params.get('state');
-
-        if (sessionState && code && state) {
-            this.userService.oidcLogin({sessionState, code, state}).pipe(first()).subscribe();
-        }
-    }
```
ChristianBeilschmidt, line 261: Why is this gone?

ChristianBeilschmidt (changes requested):

projects/core/src/lib/users/user.service.ts
michaelmattig: no. This methods checks the URL parameters which do not have to be awaited. in case they are set, the oids login happens asynchronously and false is returned so the regular login is skipped. If the parameters are not present, the login works like before.

projects/core/src/lib/users/user.service.ts
michaelmattig: no, this works just like the login worked before. The regular login works asynchronously as weill.

projects/gis/src/app/login/login.component.scss
```diff
@@ -8,8 +8,12 @@
 }
 
 .login {
-    margin: 0 auto;
     width: 24rem;
+
+    position: absolute;
+    top: 50%;
+    left: 50%;
+    transform: translate(-50%, -50%);
```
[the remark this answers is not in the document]
michaelmattig: it's better because it works. previously the login was centered horizontally but not vertically. I am sure you can do it in a different way, but this works.

projects/gis/src/app/login/login.component.ts
michaelmattig: If the oidcInit gives us the redirect url we can activate the Oidc part of the form. What is there to subscribe?

projects/gis/src/app/main/main.component.ts
michaelmattig: this is now done in the user service. Previously, the query parameters were only ever checked AFTER the userService already tried to log in anonymously. In the case where the anonymous login was disabled this would lead to errors.

projects/core/src/lib/users/oidc/oidc.component.ts
michaelmattig: Routing it to `/` doesn't work. Nothing happens in the app. maybe because it already is at `/`?

ChristianBeilschmidt: It seems like you can force a reload on the same location on extras/options:
https://angular.io/api/router/NavigationBehaviorOptions#onSameUrlNavigation

projects/core/src/lib/users/user.service.ts
ChristianBeilschmidt: refactor like discussed in chat

projects/core/src/lib/users/user.service.ts
```diff
@@ -325,6 +330,20 @@ export class UserService {
                 this.sessionQuota$.next(quota);
             });
     }
+
+    private checkOidcParametersAndLogin(): boolean {
+        const params = new URLSearchParams(window.location.search);
+        const sessionState = params.get('session_state');
+        const code = params.get('code');
+        const state = params.get('state');
```
ChristianBeilschmidt, line 338: cf. chat, removal of parameters

ChristianBeilschmidt (changes requested):

projects/core/src/lib/users/user.service.ts
michaelmattig: add09e802673f1b7d52dc319be3caa6f2b1e451e

ChristianBeilschmidt:
bors merge

ChristianBeilschmidt (approved):

ChristianBeilschmidt:
bors merge

PR #676, opened 2023-08-31
fix gradient: color between lower and higher value

jdroenner:


jdroenner:
[Vector Colorizer Linear Gradient nutzt nur die erstem beiden Einträge](https://trello.com/c/e3OoqCzH/1460-vector-colorizer-linear-gradient-nutzt-nur-die-erstem-beiden-eintr%C3%A4ge)

projects/core/src/lib/colors/colorizer.model.ts
```diff
@@ -96,13 +96,16 @@ export class LinearGradient extends Colorizer {
             return this.overColor;
         }
 
-        const index = this.breakpoints.findIndex((b) => value >= b.value);
+        const index = this.breakpoints.findIndex((b, i, brks) => {
+            return value >= b.value && (i === brks.length - 1 || value < brks[i + 1].value);
+        });
```
ChristianBeilschmidt, line 101: Can't this be done by using `findLastIndex`?

https://jsfiddle.net/8yushjno/

ChristianBeilschmidt (changes requested):

projects/core/src/lib/colors/colorizer.model.ts
ChristianBeilschmidt: We shouldn't increase the javascript version to ES2023 if this is not recommended by Angular.

Maybe just reverse the array?

https://jsfiddle.net/m38y5416/

ChristianBeilschmidt (approved):

PR #677, opened 2023-09-14
Color editing possible

ClockworkOnion:
Color palette tabs don't get updated on "Create Color Table", only after closing and re-opening the dialog or switching to another colorizer and back.
`raster-gradient-symbology-editor.component.ts` line 236, `createColorTable` seems like the best option to fix it, but I had no success so far.

### DIFF
--- a/projects/core/src/lib/colors/color-table-editor/color-table-editor.component.ts
+++ b/projects/core/src/lib/colors/color-table-editor/color-table-editor.component.ts
@@ -24,9 +24,14 @@ export class ColorTableEditorComponent implements OnInit {
     constructor(private ref: ChangeDetectorRef) {}
 
     ngOnInit(): void {
+        this.updateColorAttributes();
+    }
+
+    updateColorAttributes(): void {
         this.colorAttributes = this.colorTable.map((color: ColorBreakpoint) => {
             return {key: color.value.toString(), value: color.color};
         });
+        this.ref.detectChanges();
     }
 
     /**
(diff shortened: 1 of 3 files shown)


ClockworkOnion:
[Color Tables (Raster)](https://trello.com/c/UskOQTY9/1104-color-tables-raster)

jdroenner:
The issue is that the color table component only acts "onInit". You need to add OnChanges to let it also update when the input changes.

```
export class ColorTableEditorComponent implements OnInit, OnChanges {
...

    ngOnInit(): void {
        this.updateColorAttributs();
    }

    ngOnChanges(changes: SimpleChanges): void {
        if (changes['colorTable']) {
            this.updateColorAttributs();
        }
    }

    updateColorAttributs(): void {
        this.colorAttributes = this.colorTable.map((color: ColorBreakpoint) => {
            return {key: color.value.toString(), value: color.color};
        });
...
}
```

ClockworkOnion:
> The issue is that the color table component only acts "onInit". You need to add OnChanges to let it also update when the input changes.
> 
> ```
> export class ColorTableEditorComponent implements OnInit, OnChanges {
> ...
> 
>     ngOnInit(): void {
>         this.updateColorAttributs();
>     }
> 
>     ngOnChanges(changes: SimpleChanges): void {
>         if (changes['colorTable']) {
>             this.updateColorAttributs();
>         }
>     }
> 
>     updateColorAttributs(): void {
>         this.colorAttributes = this.colorTable.map((color: ColorBreakpoint) => {
>             return {key: color.value.toString(), value: color.color};
>         });
> ...
> }
> ```

I think this is the right direction, but updating  `colorAttributes` like this in `ngOnChanges` will cause `ngOnChanges` to call itself in an infinite loop.
I tried only calling `updateColorAttributes` from the parent component on pressing `Create Color Table`, as well as adding `detectChanges()` calls everywhere, but that didn't work either...

jdroenner (approved):

PR #679, opened 2023-10-16
show class number and names

jdroenner:


jdroenner:
[Pixel-Werte in der Color-Table anzeigen, wenn die Unit eine Classification ist](https://trello.com/c/1RnXTFBd/1397-pixel-werte-in-der-color-table-anzeigen-wenn-die-unit-eine-classification-ist)

projects/core/src/lib/layers/legend/legend-raster/raster-legend.component.html
```diff
@@ -41,6 +41,7 @@
                     <td class="classes">
                         <div class="icon" [style.background-color]="breakpoint | breakpointToCssStringPipe"></div>
                     </td>
+                    <td class="classes">{{ breakpoint.value }}</td>
```
michaelmattig, line 44: is the value or the class label more important to the user?
The value could e.g. be very large and then the label is very far apart from the color.
Maybe show the label first and then the value in parentheses?
If you like it better the way it is just keep it.

projects/core/src/lib/colors/color-attribute-input/color-attribute-input.component.html
```diff
@@ -8,6 +8,7 @@
             [ngModel]="input?.key"
             (ngModelChange)="updateKey($event)"
         />
+        <mat-hint *ngIf="hasColorHint()" align="start">{{ colorHint(input?.key) }}</mat-hint>
```
michaelmattig, line 11: what I don't really understand: The user can change the value of the breakpoint here right? Then the label/hint would no longer correspond to the class right? Isn't that a problem?

projects/core/src/lib/layers/measurement.ts
```diff
@@ -82,4 +83,9 @@ export class ClassificationMeasurement extends Measurement {
             classes: this.classes.toObject(),
         };
     }
+
+    colorHint(key: string): string | undefined {
```
michaelmattig, line 87: why do you call it color hint? It's not a color, so this may be confusing. Why not call it class label?

projects/core/src/lib/layers/legend/legend-raster/raster-legend.component.html
jdroenner: well i guess we should try it like this. if we see a classification with a number larger than 10k we can still change it ...

projects/core/src/lib/colors/color-attribute-input/color-attribute-input.component.html
jdroenner: i believe in the power of data binding here. The model changes and then the hint will also change. At least it did when i tried it.

projects/core/src/lib/layers/measurement.ts
jdroenner: because it is related to the color selection. So we might add a hint for min max values or something like that. It could be used for other colorizers then classifications.

michaelmattig (approved):

PR #680, opened 2023-10-17
add grays, enhance breakpoint logic

jdroenner:


jdroenner:
[Schwarz-Weiß Colorizer für Raster hinzufügen](https://trello.com/c/OIndNZDT/1439-schwarz-wei%C3%9F-colorizer-f%C3%BCr-raster-hinzuf%C3%BCgen)

projects/core/src/lib/colors/color-map-selector/mpl-colormaps.ts
```diff
@@ -11,12 +11,12 @@ import {RgbaTuple} from '../color';
 /**
  * The color map names.
  */
-export type MplColormapName = 'MAGMA' | 'INFERNO' | 'PLASMA' | 'VIRIDIS';
+export type MplColormapName = 'MAGMA' | 'INFERNO' | 'PLASMA' | 'VIRIDIS' | 'GRAYS';
```
michaelmattig, line 14: is "grays" the correct name in mpl? I only find "gray". Also they are all in lower case in mpl 🤔

projects/core/src/lib/colors/color-map-selector/color-map-selector.component.ts
```diff
@@ -211,17 +211,36 @@ export class ColorMapSelectorComponent implements OnInit, OnDestroy, OnChanges {
...
+                nextMapIndex = Math.max(0, colorMapIndex - 1);
+            }
+
+            if (colorMapPosRemainder < Number.EPSILON) {
+                breakpoints.push(new ColorBreakpoint(value, Color.fromRgbaLike(colorMap[colorMapIndex])));
+                continue;
             }
+
             const color = Color.fromRgbaLike(colorMap[colorMapIndex]);
+            const nextColor = Color.fromRgbaLike(colorMap[nextMapIndex]);
```
michaelmattig, line 234: maybe add a comment here that you interpolate between two steps?

why wasn't this done here before anyway?

also: does you method always work? It seems the values can go beyond 256?

maybe this would be something that could even be checked in a unit test.

projects/core/src/lib/colors/color-map-selector/color-map-selector.component.ts
```diff
@@ -211,17 +211,36 @@ export class ColorMapSelectorComponent implements OnInit, OnDestroy, OnChanges {
...
```
michaelmattig, line 235: putting this comment here randomly:

how can I try this out with the default datasets? I tried using it on land cover and ndvi but they are both palettes and if I change from palette to linear colorizer nothing, create the breakpoints and apply them, nothing happens.

projects/core/src/lib/colors/color-map-selector/mpl-colormaps.ts
jdroenner: i think it is "Grays" in the MPL source: https://github.com/matplotlib/matplotlib/blob/main/lib/matplotlib/_cm.py

projects/core/src/lib/colors/color-map-selector/color-map-selector.component.ts
jdroenner: well there was a problem when the colors are not truncated because the backend can't handle float rgba values.
The method schould always work since the frac is still the same as before just computed different. The "value" is the value of the pixel and this can be larger then 255.

jdroenner: [d31466b](https://github.com/geo-engine/geoengine-ui/pull/680/commits/d31466ba837ef2fb26d0b6740bdfbc7b47c3cf37)

projects/core/src/lib/colors/color-map-selector/color-map-selector.component.ts
jdroenner: there was this issue with the float rgba values. I fixed that: [d31466b](https://github.com/geo-engine/geoengine-ui/pull/680/commits/d31466ba837ef2fb26d0b6740bdfbc7b47c3cf37)

Also i fixed the switch between linear and log scale: [df92651](https://github.com/geo-engine/geoengine-ui/pull/680/commits/df926516b3d3cf98d814bbc58f984ea50d1d5e7d)

I always use the NDVI3857 dataset. It does not use a palette.

michaelmattig: image

I tried it out, but if I select 2 steps it produces 3. Do you want to fix this here real quick or shall I create a new card?

Everything else seems to work.

jdroenner: [d5303fd](https://github.com/geo-engine/geoengine-ui/pull/680/commits/d5303fd5d104632bba8edb5c3b4d80f06e3d0169)

PR #687, opened 2023-12-11
refactor session calls

ChristianBeilschmidt:
In openapi-client, go to typescript folder and 1x `npm install && npm run build`.
Then, refer to the `dist` directory in the package.json.

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
         "dagre": "~0.8.5",
         "dagre-d3": "~0.6.4",
         "flatgeobuf": "3.25.0",
-        "geoengine-openapi-client": "file:/home/beilschmidt/git/geo-engine/openapi-client/typescript/dist",
+        "@geoengine/openapi-client": "0.0.2",
         "immutable": "~4.3.0",
         "keycharm": "0.4.0",
         "moment": "~2.29.4",
--- a/package.json
+++ b/package.json
@@ -52,6 +52,7 @@
         "dagre": "~0.8.5",
         "dagre-d3": "~0.6.4",
         "flatgeobuf": "3.25.0",
+        "geoengine-openapi-client": "file:/home/beilschmidt/git/geo-engine/openapi-client/typescript/dist",
         "immutable": "~4.3.0",
         "keycharm": "0.4.0",
         "moment": "~2.29.4",
--- a/projects/core/src/lib/backend/backend.model.ts
+++ b/projects/core/src/lib/backend/backend.model.ts
@@ -12,15 +12,6 @@ export interface RegistrationDict {
     id: UUID;
 }
 
-export interface SessionDict {
-    id: UUID;
-    user?: UserDict;
-    created: TimestampString;
-    validUntil: TimestampString;
-    project?: UUID;
-    view?: STRectangleDict;
-}
-
 export interface UserDict {
     id: UUID;
     email?: string;
--- a/projects/core/src/lib/users/user.service.ts
+++ b/projects/core/src/lib/users/user.service.ts
@@ -189,15 +189,12 @@ export class UserService {
      * @param credentials.password The user's password.
      * @returns `true` if the login was successful, `false` otherwise.
      */
-    login(credentials: {email: string; password: string}): Observable<Session> {
+    login(userCredentials: {email: string; password: string}): Observable<Session> {
         const result = new ReplaySubject<Session>();
 
         new SessionApi()
             .loginHandler({
-                userCredentials: {
-                    email: credentials.email,
-                    password: credentials.password,
-                },
+                userCredentials,
             })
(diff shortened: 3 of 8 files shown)


ChristianBeilschmidt:
[UI-Client-Library aus OpenAPI generieren](https://trello.com/c/Barx5fXr/1535-ui-client-library-aus-openapi-generieren)

projects/core/src/lib/backend/backend.model.ts
```diff
@@ -12,15 +12,6 @@ export interface RegistrationDict {
     id: UUID;
 }
 
-export interface SessionDict {
```
michaelmattig, line 15: Do we not need to wrap everything like we do in Python (I mean converting to and from internal types)? Our core is basically also a library.

Or do we view it as an application?

projects/core/src/lib/users/user.service.ts
```diff
@@ -190,49 +195,55 @@ export class UserService {
...
-            .loginUser(credentials)
-            .pipe(mergeMap((response) => this.createSession(response)))
-            .subscribe(
-                (session) => {
-                    this.session$.next(session);
-                    result.next(session);
+
+        new SessionApi(new Configuration({basePath: this.config.API_URL}))
+            .loginHandler({
+                userCredentials: {
```
michaelmattig, line 201: can you not just pass the credentials without unpacking?

projects/core/src/lib/users/user.service.ts
```diff
@@ -190,49 +195,55 @@ export class UserService {
...
+                    email: credentials.email,
+                    password: credentials.password,
                 },
-                (error) => result.error(error),
-                () => result.complete(),
-            );
+            })
+            .then((response) => this.sessionFromDict(response))
```
michaelmattig, line 206: why two `then`s?

projects/core/src/lib/users/user.service.ts
```diff
@@ -190,49 +195,55 @@ export class UserService {
...
     }
 
     createSessionWithToken(sessionToken: UUID): Observable<Session> {
-        return this.backend.getSession(sessionToken).pipe(
-            mergeMap((response) => this.createSession(response)),
-            tap((session) => this.session$.next(session)),
-        );
+        return from(
+            new SessionApi(new Configuration({basePath: this.config.API_URL, accessToken: sessionToken}))
+                .sessionHandler()
```
michaelmattig, line 244: I find these names rather confusing. I think it would be better if it's called getSessionHandler. I know it's the same in Python, so maybe we should think about renaming the handlers in Rust?

package.json
```diff
@@ -52,6 +52,7 @@
         "dagre": "~0.8.5",
         "dagre-d3": "~0.6.4",
         "flatgeobuf": "3.25.0",
+        "geoengine-openapi-client": "file:/home/beilschmidt/git/geo-engine/openapi-client/typescript/dist",
         "immutable": "~4.3.0",
```
michaelmattig, line 56: I get this warning when I run it:

> Warning: /home/michael/git/geoengine-ui/dist/core/fesm2022/geoengine-core.mjs depends on 'geoengine-openapi-client'. CommonJS or AMD dependencies can cause optimization bailouts.
For more info see: https://angular.io/guide/build#configuring-commonjs-dependencies

projects/core/src/lib/backend/backend.model.ts
ChristianBeilschmidt: Well, there is a session class that wraps it. But we can replace the dict types.

projects/core/src/lib/users/user.service.ts
ChristianBeilschmidt: Ok, we should discuss it, but not in this task.

package.json
ChristianBeilschmidt: We needed to add `/esm` to the end.
The current package works fine: [43cddc1](https://github.com/geo-engine/geoengine-ui/pull/687/commits/43cddc17438a148c50093ad2371ff3f3485dbf1f)

projects/core/src/lib/users/user.service.ts
ChristianBeilschmidt: [794cdc0](https://github.com/geo-engine/geoengine-ui/pull/687/commits/794cdc015b8891f366415fcd572c528d8d734f82)

projects/core/src/lib/users/user.service.ts
ChristianBeilschmidt: [794cdc0](https://github.com/geo-engine/geoengine-ui/pull/687/commits/794cdc015b8891f366415fcd572c528d8d734f82)

jdroenner (approved):

PR #694, opened 2024-01-15
EBV Prefill: auto fill path

ChristianBeilschmidt:


projects/dashboards/ebv-analyzer/src/app/ebv-selector/ebv-selector.component.ts
```diff
@@ -97,10 +96,21 @@ export class EbvSelectorComponent implements OnInit {
     layerSelected(id?: ProviderLayerIdDict): void {
         this.layerId = id;
 
-        if (this.autoShowEbv) {
-            this.autoShowEbv = false;
-            this.showEbv();
+        this.showEbv();
+    }
+
+    pathChange(collections: LayerCollectionDict[]): void {
```
michaelmattig, line 102: would it make sense to abort further loading the tiles once the selection changes?

At least for me it's lagging heavily behind, when I change the layer while it is still loading. But may be because of the developer mode with limited requests.

michaelmattig (approved):

PR #710, opened 2024-03-07
Input layer removal checkbox

dbrandenstein:


projects/core/src/lib/layers/layer-list/layer-list-element/layer-list-element.component.html
```diff
@@ -24,7 +24,7 @@
                 <mat-icon>merge_type</mat-icon>
                 <span class="mat-body-2">Lineage</span>
             </button>
-            <button mat-menu-item (click)="projectService.removeLayer(layer)">
+            <button mat-menu-item (click)="projectService.removeLayer(layer).subscribe()">
```
ChristianBeilschmidt, line 27: Why is it necessary to subscribe here when it wasn't before?

projects/core/src/lib/operators/dialogs/helpers/layer-selection/layer-selection.component.html
```diff
@@ -13,6 +13,7 @@ <h4>Select input {{ title }}</h4>
     </button>
 </div>
 <geoengine-layer-collection-layer-details [layerMetadata]="metadata" *ngIf="expanded"></geoengine-layer-collection-layer-details>
+<mat-checkbox (click)="toggleRemoveAfterCreation()">Remove after operator creation</mat-checkbox>
```
ChristianBeilschmidt, line 16: grafik

I'm not quite sure that providing multiple checkboxes that (somewhat) bloat the dialog, really helps the user.

Can we come up with something more elegant?

For instance, one checkbox beside the "add" button?
Is there a need for the user to selectively remove certain inputs but not the others? I think they could do this via the layer menu.

Moreover, isn't this just a thing for the raster vector join where a user thought: Well, I have the same points than before, it could have removed the others.

So, is this a general "thing" for all operator dialogs? And if so, shouldn't it be checked for certain dialogs in the first place, and not for others?

I think we should first talk about this one. Then, I can review the solution. But my stomach tells me I'm not convinced, yet.

ChristianBeilschmidt (changes requested):

PR #720, opened 2024-03-19
Map resolution and extent info

dbrandenstein:


dbrandenstein:
[Info-Box über Auflösung und Anfragerechteck](https://trello.com/c/aIlip82E/1643-info-box-%C3%BCber-aufl%C3%B6sung-und-anfragerechteck)

projects/core/src/lib/map/map-info/map-info.component.scss
```diff
@@ -0,0 +1,12 @@
+.map-info {
+    z-index: 1;
+    position: absolute;
+    margin-bottom: 10px;
```
michaelmattig, line 4: maybe better work with rem instead of pixels

projects/dashboards/gfbio/src/app/app.component.html
```diff
@@ -65,6 +65,10 @@
             </ng-container>
         </geoengine-map-container>
     </div>
+    <geoengine-map-info
```
michaelmattig, line 68: this gets covered up if there are too many layers

michaelmattig: image

projects/core/src/lib/map/map-info/map-info.component.ts
```diff
@@ -0,0 +1,39 @@
...
+    ) {}
+
+    info(): Observable<string> {
+        return this.projectService.getSpatialReferenceStream().pipe(
+            map((srs) => this.spatialReferenceService.getOlProjection(srs).getUnits()),
+            combineLatestWith(this.mapService.getViewportSizeStream()),
+            map(
+                ([unit, viewport]) =>
+                    `Resolution: ${viewport.resolution} ${unit}/px
+                     x: (${viewport.extent[0]}, ${viewport.extent[2]})
```
michaelmattig, line 34: the text gets very long with lots of decimals. maybe only show the first two and then ... and show the real value on click?

projects/dashboards/gfbio/src/app/app.component.html
```diff
@@ -65,6 +65,10 @@
...
+        *ngIf="config.COMPONENTS.MAP_INFO.AVAILABLE"
```
michaelmattig, line 69: I wasn't really involved in the design process of this, if there was any. But: isn't this a bit too prominently places? it takes a lot of space that covers the map.

Can we not instead put this info in the top bar? We also show the selected time there, so I think it would make sense to show the bbox there as well, as both make up the query rectangle.

projects/core/src/lib/map/map-info/map-info.component.ts
```diff
@@ -0,0 +1,39 @@
...
+                     y: (${viewport.extent[1]}, ${viewport.extent[3]})`,
```
michaelmattig, line 35: maybe it would make more sense to show min and max (of x and y) on top of each other and aligned?

like

min: x: -180 y: -90
max: x:  180 y:  90

then I think it's more easily visible how large the bbox is in each dimension. But this only makes sense if the numbers are aligned though.

projects/gis/src/app/app-config.service.ts
```diff
@@ -11,6 +11,9 @@ interface Components {
     readonly REGISTRATION: {
         readonly AVAILABLE: boolean;
     };
+    readonly MAP_INFO: {
```
michaelmattig, line 14: map info is very generic. maybe call it MAP_BBOX_RESOLUTION_OVERLAY?
or something more compact but more recognizable if you have an idea.

projects/core/src/lib/map/map-info/map-info.component.scss
dbrandenstein: https://github.com/geo-engine/geoengine-ui/pull/720/commits/f9fa9764e7913450d7380b862ca6694e3092f31f

projects/dashboards/gfbio/src/app/app.component.html
dbrandenstein: https://github.com/geo-engine/geoengine-ui/pull/720/commits/4206dae5ccb78469753122cd3e8a3e24564861f8

projects/core/src/lib/map/map-info/map-info.component.ts
dbrandenstein: https://github.com/geo-engine/geoengine-ui/pull/720/commits/2ac20ad03521e0f6d9bbd9c750a252005a324b2e

projects/core/src/lib/map/map-info/map-info.component.ts
dbrandenstein: https://github.com/geo-engine/geoengine-ui/pull/720/commits/2ac20ad03521e0f6d9bbd9c750a252005a324b2e

projects/gis/src/app/app-config.service.ts
dbrandenstein: https://github.com/geo-engine/geoengine-ui/pull/720/commits/5f68e95e3c0c65f8f6ba5500915bc974b8d3fce8
https://github.com/geo-engine/geoengine-ui/pull/720/commits/10617c83eb85f4dc97e6169dd0eaa62681a39ae3

projects/core/src/lib/map/map-info/map-info.component.ts
michaelmattig: it's not enough to add a plus sign in front to properly align the numbers:

image

either we live with the numbers not being properly aligned, then we can remove the plus sign
or we properly align it. But then I guess we would need a monospace font and right-align the numbers, because then the decimal point would always be at the same place.

I'm fine with either solution.

projects/core/src/lib/map/map-info/map-resolution-extent-overlay.component.ts
```diff
@@ -0,0 +1,61 @@
...
+    private static readonly lowNumFractions = 2;
+    private static readonly highNumFractions = 12;
+
+    constructor(
+        private mapService: MapService,
+        private spatialReferenceService: SpatialReferenceService,
+        private projectService: ProjectService,
+    ) {}
+
+    togglePrecision(): void {
```
michaelmattig, line 31: can you show a hand cursor when hovering the overlay so that users know something happens on click?

dbrandenstein: https://github.com/geo-engine/geoengine-ui/pull/720/commits/fca7837d75a4889040b71ba97919799f3f9f2f77

projects/core/src/lib/map/map-info/map-info.component.ts
dbrandenstein: https://github.com/geo-engine/geoengine-ui/pull/720/commits/b5643386f00df93ec55b0986b052a1d7365aa027

michaelmattig (approved):

PR #734, opened 2024-04-19
Layer download improvements

dbrandenstein:


dbrandenstein:
[Download von Vektorlayern](https://trello.com/c/48SEXidn/1644-download-von-vektorlayern)

projects/core/src/lib/download-layer/download-layer.component.ts
```diff
@@ -132,6 +135,24 @@ export class DownloadLayerComponent implements OnInit, OnDestroy {
         this.viewportSizeSubscription?.unsubscribe();
     }
 
+    unsubscribeTime(): void {
+        this.projectTimeSubscription?.unsubscribe();
+    }
+
+    setEditedExtent(): void {
+        this.editedExtent = true;
```
michaelmattig, line 143: this should also be set when the user selected a region using "select region on the map"

projects/core/src/lib/download-layer/download-layer.component.ts
```diff
@@ -132,6 +135,24 @@ export class DownloadLayerComponent implements OnInit, OnDestroy {
...
+        this.maybeUnsubscribeViewportSize();
+    }
+
+    setEditedResolution(): void {
+        this.editedResolution = true;
```
michaelmattig, line 148: this does not work for me. If I edit the resolution and zoom in, the value changes.

dbrandenstein: Weird, pretty sure it worked before, but I can reproduce it. It now listens to the 'input' event instead of 'change' which should trigger every time the value is changed by the user.
https://github.com/geo-engine/geoengine-ui/pull/734/commits/3023343c6568b80dcc11724c531865fdaf8f6077

projects/core/src/lib/download-layer/download-layer.component.ts
dbrandenstein: https://github.com/geo-engine/geoengine-ui/pull/734/commits/54a43b93387d478e4c1ae46410c05115db03396b

michaelmattig (approved):

PR #748, opened 2024-07-03
FAIR dataset deletion in data manager

glombiewski:
Expose new backend functionality for dataset deletion in the data manager UI.


glombiewski:
[Löschen von Datasets](https://trello.com/c/06fpUSO8/1704-l%C3%B6schen-von-datasets)

PR #788, opened 2025-05-16
Measurement editing

0xmycf:
Allows for changing the `Measurements` in the `LoadingInformation` for Gdal and Ogr Sources.

## GdalSources

gdalsources

## OgrSources

ogrsources

Currently the `class values` for the `classification measurement` are typed for integers other input is not allowed.
I think this makes it impossible for declaring a proper classification measurement for `text` or `float` columns.

I did not implement this in the frontend yet, as I've seen that the backend defines the type as 

```rust
pub struct ClassificationMeasurement {
    pub measurement: String,
    pub classes: HashMap<u8, String>,
}
```

and I am not 100% sure if this should be even possible.

### Measurement Component

- The `MeasurementComponent` has been made a standalone component. I made this change, because I got a cyclic import problem with using the `MeasurementComponent` in the `OgrDatasetComponent`.
- A new `@Output() onInputChange = new EventEmitter<Event>();` that fires every time one of the text input fields changes (used to mark the form as dirty)
- `@Output() measurementChange = new EventEmitter<Measurement>();` now fires after measurement changes (didnt fire at all before)
- change `constructor` to `ngOnChange` and setup the measurement appropriately if one is given as `@Input`

### GdalMetadataListComponent and OgrDatasetComponent
- both now have a child `MeasurementComponent`

#### OgrDatasetComponent
- holds additional info about the columns in the Dataset
- tracks which dataset's measurement was edited last and which is being currently edited (to properly save the new measurement`
- now implements `OnInit` to setup the currently selected column for the measurement component

### DIFF
--- a/projects/common/src/lib/datasets/ogr-dataset/ogr-dataset.component.html
+++ b/projects/common/src/lib/datasets/ogr-dataset/ogr-dataset.component.html
@@ -205,4 +205,25 @@
         <mat-label>Spatial Reference</mat-label>
         <input matInput type="text" formControlName="spatialReference" />
     </mat-form-field>
+
+    <mat-form-field class="column">
+        <mat-label>Column</mat-label>
+        <mat-select [ngModel]="selectedColumn()" (ngModelChange)="setSelectedColumn($event)" [ngModelOptions]="{standalone:true}"
+                    (selectionChange)="onColumnMatSelectChange($event)">
+            <!--
+            this causes an infinite loop in.
+            I think that's because we create a new column vector every time we call this method?
+            -->
+            <!--            <mat-option *ngFor="let col of _columns" [value]="col">{{ col }}</mat-option>-->
+            @for (col of sortedColumns(); track col.name) {
+                <mat-option [value]="col.name">{{ col.name }}</mat-option>
+            }
+        </mat-select>
+    </mat-form-field>
+    <geoengine-measurement
+        [measurement]="newMeasurementForColumn()"
+        (measurementChange)="onMeasurementChange($event)"
+        (onInputChange)="markDirty()"
+    ></geoengine-measurement>
+
 </form>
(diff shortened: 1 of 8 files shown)


dbrandenstein:
[Measurement Editing](https://trello.com/c/T5m4auZs/1839-measurement-editing)

projects/common/src/lib/datasets/ogr-dataset/ogr-dataset.component.html
```diff
@@ -205,4 +205,28 @@
         <mat-label>Spatial Reference</mat-label>
         <input matInput type="text" formControlName="spatialReference" />
     </mat-form-field>
+
+    <mat-form-field class="column">
```
dbrandenstein, line 209: Could you include some sort of indication in the UI that this section is meant for setting the measurement? Alternatively, would it maybe make sense to have the "geoengine-measurement" appear directly below the "Integer Columns" etc. fields when one clicks on a column name/enters a new one there? It might be more intuitive to have it directly there, associated with the column definition and would be neat to have it only appear when actively editing the column names this way and hidden otherwise.

projects/manager/src/app/datasets/loading-info/gdal-metadata-list/gdal-metadata-list.component.html
```diff
@@ -31,6 +31,13 @@
             <mat-label>Band Name</mat-label>
             <input matInput type="text" formControlName="bandName" />
         </mat-form-field>
+
+        <ng-container *ngIf="metaData as meta">
+            <geoengine-measurement
```
dbrandenstein, line 36: I think it would be good to add an indication that the measurement is set in this section

0xmycf: https://github.com/geo-engine/geoengine-ui/pull/788/commits/0dee455e72e64d482bd8c8118dc65ce5edafb606

PR #789, opened 2025-05-21
fix(manager): save NoDataValue in `gdal-dataset-parameters` correctly

0xmycf:
The `noDataValue` was not bound to the form field for "No Data Value", instead it was bound to height in `gdal-dataset-parameters.component.html`.


dbrandenstein (approved):

dbrandenstein:
[fix(manager): save NoDataValue in gdal-dataset-parameters correctly](https://trello.com/c/KXhrWh44/1840-fixmanager-save-nodatavalue-in-gdal-dataset-parameters-correctly)

PR #800, opened 2025-07-28
refactor: Injectors instead of constructor parameters

ChristianBeilschmidt:


eslint.config.mjs
```diff
@@ -63,6 +63,7 @@ export default defineConfig([
                     style: 'camelCase',
                 },
             ],
+            '@angular-eslint/prefer-signals': 'warn',
```
michaelmattig, line 66: unrelated to this line: is there a reason you did not use the inject here?

alt text

ChristianBeilschmidt: These are no components, thus there is no dependency injection.

michaelmattig (approved):

